PR #9, opened 2017-06-24
Implement ACKs

ekr:
I have taken a slightly different approach here than we discussed
in Chicago. Specifically:

- ACKs are not handshake messages but a new content type. This
  avoids the need to ack ACKs or worry about reassembling
  when ACKs are lost.

- The receipt of a flight also counts as an implicit ACK and
  so you shouldn't ACK full flights. I tried removing this
  feature but it leads to weird anomalies where you have
  two flights you should be retransmitting if the ACKs get
  lost.

- ACKs are always encrypted with the then-current key.

- I also removed the rule that you don't stop retransmitting
  when you receive a partial flight. That was intended to have
  fast retransmission for missing messages because the retransmit
  of flight A triggered retransmission of partial flight B, but
  ACKs serve this purpose.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -532,7 +532,7 @@ the following changes:
 
 3. The TLS 1.3 KeyUpdate message is not used in DTLS 1.3 for re-keying.
 
-4. A new ACK message has been added for reliable message delivery of certain handshake messages.
+4. A new ACK content type has been added for reliable message delivery of handshake messages.    
 
 Note that TLS 1.3 already supports a cookie extension, which used to
 prevent denial-of-service attacks. This DoS prevention mechanism is
@@ -720,10 +720,27 @@ such as version, random, ciphersuites. The server MUST use the same
           case finished:              Finished;
           case new_session_ticket:    NewSessionTicket;
           case key_update:            KeyUpdate; /* reserved */
-          case ack:                   ACK; /* DTLS-required field */
       } body;
   } Handshake;
 ~~~~
+
+The first message each side transmits in each association always has
+message_seq = 0.  Whenever a new message is generated, the
+message_seq value is incremented by one. When a message is
+retransmitted, the old message_seq value is re-used, i.e., not 
+incremented. From the perspective of the DTLS record layer, the retransmission is
+a new record.  This record will have a new
+DTLSPlaintext.sequence_number value.
+
+
+DTLS implementations maintain (at least notionally) a
+next_receive_seq counter.  This counter is initially set to zero.
+When a message is received, if its sequence number matches
+next_receive_seq, next_receive_seq is incremented and the message is
+processed.  If the sequence number is less than next_receive_seq, the
+message MUST be discarded.  If the sequence number is greater than
+next_receive_seq, the implementation SHOULD queue the message but MAY
+discard it.  (This is a simple space/bandwidth tradeoff).
 
 In addition to the handshake messages that are deprecated by the TLS 1.3
 specification DTLS 1.3 furthermore deprecates the HelloVerifyRequest message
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1350,6 +1323,83 @@ ClientHello                 -------->
...
+    two circumstances: 
+
+    - When they receive a message or fragment which is out of order,
+    either because it is not the next expected message or because
+    it is not the next piece of the current message.
+
+    - When they have received part of a flight and do not immediately
+    receive the rest of the flight (which may be in the same UDP
+    datagram). A reasonable approach here is to
+    set a timer for 1/4 the current retransmit timer value when
```
ianswett: Is the current retransmit timer a single RTT?  Commonly the retransmission timer(ie: RTO timer) is based on the delayed ack time, so I think basing it on smoothed RTT or another RTT based metric is wiser than specifying it from a retransmit timer, assuming that's possible.

1/4 RTT seems like a good recommendation for most RTTs, though for very small RTTs, this could be very small, particularly if it's smaller than the timer granularity.  I'm not sure you need explicit wording on that or not?

ekr: It's actually pretty much undefined. We mostly just say "start with a timer of X and then back off".  For the main handshake, generally the only flight that is sent with an RTT estimate is the client's second flight, so it's kind of ad hoc. @ianswett would welcome wording here.

draft-ietf-tls-dtls13.md
```diff
@@ -1350,6 +1323,83 @@ ClientHello                 -------->
...
+
+    When an implementation receives a partial flight, it SHOULD generate
+    an ACK that covers the messages from that flight which it has
+    received so far. Implementations have some discretion about when
+    to generate ACKs, but it is RECOMMENDED that they do so under
+    two circumstances: 
+
+    - When they receive a message or fragment which is out of order,
+    either because it is not the next expected message or because
+    it is not the next piece of the current message.
```
ekr: Good question. TCP does this, but then we're not TCP

draft-ietf-tls-dtls13.md
```diff
@@ -1117,32 +1022,44 @@ timeout and retransmission calculation.
...
+   remains in WAITING, or, if the ACK was for the final flight,
+   transitions to FINISHED.
+
+3. The implementation reads a retransmitted flight from the peer: the
    implementation transitions to the SENDING state, where it
    retransmits the flight, resets the retransmit timer, and returns
    to the WAITING state.  The rationale here is that the receipt of a
    duplicate message is the likely result of timer expiry on the peer
    and therefore suggests that part of one's previous flight was
    lost.
```
ekr: It actually is mentioned but I added an xref

draft-ietf-tls-dtls13.md
```diff
@@ -1350,6 +1323,83 @@ ClientHello                 -------->
 ~~~~
 {: #dtls-msg-epoch title="Example DTLS Exchange with Epoch Information"}
 
+
+# ACK Message {#ack-msg}
+
+The ACK handshake message is used by an endpoint to indicate the
```
ekr: That's just a leftover. It's not ahandshake message.

draft-ietf-tls-dtls13.md
```diff
@@ -1350,6 +1323,83 @@ ClientHello                 -------->
...
+TLS records it has received from the other side. ACK is not
+a handshake message but is rather a separate content type,
+with code point TBD. This avoids it consuming space in the
+handshake message sequence. Note that ACKs can still be
+piggybacked on the same UDP datagram as handshake records.
```
ekr: That's below

PR #16, opened 2017-10-26
Short header format

ekr:
@martinthomson, PTAL

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -306,8 +306,11 @@ length:
 
 encrypted_record:
 : Identical to the encrypted_record field in a TLS 1.3 record.
-
 {:br}
+
+As with previous versions of DTLS, mutliple DTLSCiphertext records can be included
+in the same underlying transport datagram.
+
 
 The short DTLS header format is:
 
@@ -341,7 +344,8 @@ DTLSShortCiphertext format records in the same datagram.
 
 DTLSShortCiphertext MUST only be used for data which is protected with
 one of the application_traffic_secrets, and not for either
-handshake or early data. When using the application_traffic_secret,
+handshake or early data. When using an application_traffic_secret
+for message protection,
 Implementations MAY use either DTLSCiphertext or DTLSShortCiphertext
 at their discretion.
 
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -258,10 +258,17 @@ modify their data transmission strategy.
 
 # The DTLS Record Layer
 
-The DTLS record layer is similar to that of TLS 1.3 unless noted otherwise.
-The only change is the inclusion of an explicit epoch and sequence number
+The DTLS record layer is similar to that of TLS 1.3.
+There are two major changes:
+
+1. The DTLS record layer includes an explicit epoch and sequence number
 in the record.  This sequence number allows the recipient to correctly
-verify the TLS MAC.  The DTLS record format is shown below:
+verify the TLS MAC.
+
+2. DTLS adds a short header format that can be used to reduce overhead
+once the handshake is complete.
+
+The basic DTLS record format is shown below:
 
 ~~~~
   struct {
@@ -277,7 +284,7 @@ verify the TLS MAC.  The DTLS record format is shown below:
        uint48 sequence_number;               // DTLS-related field
        uint16 length;
        opaque encrypted_record[length];
-} DTLSCiphertext;
+    } DTLSCiphertext;
 ~~~~
 
 type:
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -300,6 +307,44 @@ length:
...
+    } DTLSShortCiphertext;
+~~~~
+
+The short_epoch_and_sequence document contains the epoch and sequence
+packed into a 16 bit integer as follows:
+
+- The first three bits are set to 001 in order to allow demuxing
+  between DTLS and other protocols {{?RFC7983}} and distinguish the
+  short from long header formats.
+
```
hannestschofenig: maybe it should say "allow multiplexing between DTLS and VoIP protocols (STUN, RTP/RTCP, etc.)

draft-ietf-tls-dtls13.md
```diff
@@ -347,6 +392,51 @@ allowing the sequence number to wrap.
 Implementations MUST NOT allow the epoch to wrap, but instead MUST establish
 a new association, terminating the old association.
 
+### Determining the Header Format
+
+Implementations can distinguish the two header formats by examining
+the first byte, which in the DTLSCiphertext header represents the
+content type. The only valid values for the content type are
+alert(21), handshake(22), or application_data(23), which actually
```
hannestschofenig: there is also the ACK message, which has a separate content type.

ekr:
Oops. Thanks.

On Thu, Oct 26, 2017 at 1:11 PM, Hannes Tschofenig <notifications@github.com

draft-ietf-tls-dtls13.md
```diff
@@ -300,6 +307,44 @@ length:
...
+- The first three bits are set to 001 in order to allow multiplexing
+  between DTLS and VoIP protocols (STUN, RTP/RTCP, etc.) {{?RFC7983}}
+  and distinguish the short from long header formats.
+
+- The fourth bit is the low order bit of the epoch value.
+
+- The remaining bits contain the low order 12 bits of the sequence
+  number.
+
+For deprotecion purposes, the nonce is computed using the full
```
martinthomson: deprotection doesn't appear to be a word, maybe "when removing packet protection, the nonce is computed..."

draft-ietf-tls-dtls13.md
```diff
@@ -300,6 +307,44 @@ length:
 encrypted_record:
 : Identical to the encrypted_record field in a TLS 1.3 record.
 
+{:br}
+
```
martinthomson: At this point I would point out that you can include multiple records in the same datagram.

draft-ietf-tls-dtls13.md
```diff
@@ -347,6 +392,52 @@ allowing the sequence number to wrap.
...
+with application_data actually
+representing protected data with the true content type being
+encrypted. If any of these values is present, then the record MUST be
+handled as DTLSCiphertext.
+
+If the first byte is any other other value, then receivers
+MUST check to see if the leading bits of the first byte are
+001. If so, they MUST process the record as DTLSShortCiphertext.
+Otherwise, the record MUST be rejected as if it had failed
+deprotection.
```
martinthomson: deprotection again

draft-ietf-tls-dtls13.md
```diff
@@ -347,6 +392,52 @@ allowing the sequence number to wrap.
...
+If the epoch bit has the same parity as the current epoch, then
+implementations SHOULD reconstruct the sequence number by computing
+the full sequence number which is numerically closest to the most
+recently successfully deprotected record.
+
+If the epoch bit has a different parity from the current epoch, then
+the record is either from a previous epoch or from a future
+epoch. Implementations SHOULD use the epoch value which would produce
+a sequence number which is numerically closest, as determined by the
+algorithm in the paragraph above.
```
martinthomson: Are you suggesting that the next and previous epoch are checked?  Or maybe all potentially active epochs?  If so, you need to mention that the next sequence number is from the selected epoch and not the current one.

I think that we can avoid some of the potential for variation with a different algorithm (noting of course that you permit variations).  If we restrict KeyUpdate to only one outstanding change, then you only ever have to select between two epochs.  If you require acknowledgment of KeyUpdate before switching epochs, then you know to expect a change.

Then, without KeyUpdate, you have just one epoch and can discard records that don't match parity.

With KeyUpdate (which you should put in a new section), you have two epochs active from the moment that you send an acknowledgment for a KeyUpdate.  You can select between them using the bit.

ekr: I agree with these comments. Note the OPEN ISSUE markers below. I was hoping to just get this in.

ekr:
Yes, I thought about shrinking the other format, but I think these are orthogonal, so I figured I'd hold off.

draft-ietf-tls-dtls13.md
```diff
@@ -347,6 +392,52 @@ allowing the sequence number to wrap.
...
+are used to compute the per-record nonce, failure to reconstruct these
+values leads to failure to deprotect the record, and so implementations
+MAY use a mechanism of their choice to determine the full values.
+This section provides an algorithm which is comparatively simple
+and which implementations are RECOMMENDED to follow.
+
+If the epoch bit has the same parity as the current epoch, then
+implementations SHOULD reconstruct the sequence number by computing
+the full sequence number which is numerically closest to the most
+recently successfully deprotected record.
```
martinthomson: Missed this:  it isn't "most recently".  In QUIC we use the numeric closest to the highest successfully decrypted packet plus one.  Or the closest to the next expected packet.  Otherwise delayed packets will shift the window backwards.  (Basically, this means that negative differences are used to allow for delays, and positive differences for lost packets).

draft-ietf-tls-dtls13.md
```diff
@@ -300,6 +307,44 @@ length:
...
+{{reconstructing}}.
+
+In this format, the length field is omitted and therefore the
+record consumes the entire rest of the datagram in the lower
+level transport. It is not possible to have multiple
+DTLSShortCiphertext format records in the same datagram.
+
+DTLSShortCiphertext MUST only be used for data which is protected with
+one of the application_traffic_secrets, and not for either
+handshake or early data. When using the application_traffic_secret,
```
kaduk: Maybe "an application_traffic_secret for message protection" to be consistent with the previous sentence?

PR #17, opened 2017-10-26
Reformatted Transport Mapping Section

hannestschofenig:
Added references to RSL/MFL.
Re-arranged the text that belongs together.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -61,6 +61,8 @@ informative:
   RFC5246:
   RFC6347:
   RFC7525:
+  I-D.ietf-tls-record-limit:
+  RFC6066:
 --- abstract
 
 This document specifies Version 1.3 of the Datagram Transport Layer Security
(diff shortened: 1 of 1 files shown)


hannestschofenig:
There is one sentence from version -01 that appears to be wrong: 
"
   Multiple DTLS records may be placed in a single datagram.  They are
   simply encoded consecutively.  The DTLS record framing is sufficient
   to determine the boundaries.  Note, however, that the first byte of
   the datagram payload must be the beginning of a record.  Records may
   not span datagrams.
"

When the first byte of a datagram must be the beginning of a record then it is not possible for a record to span multiple datagrams. Hence, the sentence "Records may not span datagrams." should actually read "Records MUST NOT span datagrams."

draft-ietf-tls-dtls13.md
```diff
@@ -347,42 +347,27 @@ allowing the sequence number to wrap.
...
-immediately, potentially leading to timeouts and spurious
-retransmission.  When DTLS is used over such transports, care should
-be taken not to overrun the likely congestion window. {{RFC5238}}
-defines a mapping of DTLS to DCCP that takes these issues into account.
-
-
-##  PMTU Issues
+- Multiple DTLS records may be placed in a single datagram. They are
+simply encoded consecutively. The DTLS record framing is sufficient
+to determine the boundaries.
```
hannestschofenig: It might be worthwhile to reference the additional rules related to the record layer definition in Section 5 of draft-ietf-tls-tls13.

ekr:
@hannestschofenig you seem to have reverted some other people's commits. Can you rebase?

draft-ietf-tls-dtls13.md
ekr: Actually, I think this text is right. The first byte must be the beginning of a record but not vice versa.

ekr:
This definitely isn't going to apply any more. I'm closing it. Hannes, feel free to resubmit with an updated version,

PR #20, opened 2017-10-26
Shrink long header

ekr:


draft-ietf-tls-dtls13.md
```diff
@@ -272,34 +281,62 @@ verify the TLS MAC.  The DTLS record format is shown below:
...
-
-epoch:
-: A counter value that is incremented on every cipher state change.
-
-sequence_number:
-: The sequence number for this record.
+epoch_and_sequence:
+: The low order two bits of the epoch and the low order 30 bits of
+  the sequence number, laid out as a 32 bit integer with the
+  epoch constituting the first two bits and the sequence number
```
martinthomson: "The first 2 bits hold low order bits from the epoch and the remaining 30 bits hold the low order bits from the sequence number."  Or something like that.

draft-ietf-tls-dtls13.md
```diff
@@ -347,6 +384,53 @@ allowing the sequence number to wrap.
...
+MAY use a mechanism of their choice to determine the full values.
+This section provides an algorithm which is comparatively simple
+and which implementations are RECOMMENDED to follow.
+
+If the epoch bits match those of the current epoch, then
+implementations SHOULD reconstruct the sequence number by computing
+the full sequence number which is numerically closest to one plus the
+sequence number of the highest successfully deprotected record.
+
+If the epoch bit has a different parity from the current epoch, then
```
martinthomson: The epoch is two bits in the long header.

ekr:
Merged by hand

PR #21, opened 2017-10-27
Move back to a KeyUpdate-type structure for updating keys. This works

ekr:
better now that we have ACK. This PR is technically independent of PR#20,
but actually has a dependency because of epoch reconstruction. Before
landing I'll rebase and fix up the reference.


ekr:
Merged by hand

Issue #56, opened 2018-08-24
New header format: proposal to reduce the amount of code points implicitly consumed


thomas-fossati:
Instead of `001`, make the fixed prefix `0011` by picking one of the three reserved bits:
```
 +-+-+-+-+-+-+-+-+
 |0|0|1|1|C|L|X|X|
 +-+-+-+-+-+-+-+-+
 |Ep.| 14 bit    |   Legend:
 +-+-+           |
 |Sequence Number|   Ep. - Epoch
 +-+-+-+-+-+-+-+-+   C   - CID present
 | Connection ID |   L   - Length present
 | (if any,      |   X   - Reserved
 /  length as    /
 |  negotiated)  |
 +-+-+-+-+-+-+-+-+
 | 16 bit Length |
 | (if present)  |
 +-+-+-+-+-+-+-+-+
```
This way the implicit Content Type allocation is halved to [48, 63] and we get back 16 codepoints.

martinthomson:
I think that we can do better.  I've just opened #58.  That uses all the bits in that first octet, which would seem to be a regression, but there are two ways in which we could get another bit back.

The first option is to drop the S-bit I propose and have sequence numbers be 16 bits always.  I'd be sad that we lost the small optimization, but that isn't the end of the world.

Perhaps a better option is to drop the second epoch bit.  This seems like a bad idea because you potentially have multiple epochs inbound to a server during a 0-RTT handshake.  In that case, you have to distinguish between ClientHello (001 or 0011 would work for that), 0-RTT (epoch 1), Handshake (epoch 2), and the ultimate Application Data (epoch 3).  If you only have one bit of epoch, then 1 and 3 are indistinguishable.

However, if you look at the way we handle 0-RTT in stacks, it might be OK to allow those late-arriving 0-RTT records to be dropped.  For instance, NSS processes 0-RTT and handshake linearly.  We don't read the Finished until we have the EndOfEarlyData, and as soon as we process the EOED, we stop accepting new 0-RTT.  Part of the reason for this is that once we report handshake success, we don't provide any other signal to an application about the data it receives.  So, for NSS, dropping 0-RTT when 1-RTT is available wouldn't change anything for us.

[...]

PR #58, opened 2018-08-31
Don't bother reserving bits

martinthomson:
This is a bikeshed colour, but I think a good one.  In the 5 bits that
we have, we can pack two epoch bits (we only need one in practice, but
let's leave that aside for the moment), and a bit each to signal
presence of connection ID and length.  That leaves one spare bit.  I'm
suggesting that we spend this on signaling the difference between an
8-bit and 16-bit sequence number.  16 bits is far more comfortable than
14, and 8 is probably enough for many use cases (saving a byte).

That is, unless you were saving this for a spin bit and VEC.

This doesn't do anything about #56, but I think that we can save a bit
from the epoch.  More on that later.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -319,15 +319,14 @@ The DTLSCiphertext header is tightly bit-packed, as shown below:
 ~~~~
   0 1 2 3 4 5 6 7
  +-+-+-+-+-+-+-+-+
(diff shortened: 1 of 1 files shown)


thomas-fossati:
If epoch can be one bit - the [rationale](https://github.com/tlswg/dtls13-spec/issues/56#issuecomment-417561065) you provided is quite convincing - then this proposal could coexist with #56.

PR #62, opened 2018-10-22
Remove EOED. This aligns with QUIC.

ekr:


ekr:
@martinthomson

martinthomson:
s'OK wi' me.

The other thing to consider mentioning is that record shadowing (key compromise at epoch N allowing the attacker to send records that cause records at epoch >N to be dropped) isn't a concern because dropping records in DTLS is trivial.

ekr:
@adl: thoughts?

draft-ietf-tls-dtls13.md
```diff
@@ -986,6 +986,19 @@ part of the same flight.  Thus, there are two acceptable ways to pack
...
 
+##  End Of Early Data
+
+The DTLS 1.3 handshake has one important difference from the
+TLS 1.3 handshake: the EndOfEarlyData message is omitted both
+from the wire and the handshake transcript: because DTLS
+records have epochs, EndOfEarlyData is not necessary to determine
+when the early data is complete, and because DTLS is lossy,
+attackers can trivially mount the deletion attacks that EndOfEarlyData
+prvents in TLS. Servers SHOULD aggressively
```
martinthomson:
```suggestion
prevents in TLS. Servers SHOULD aggressively
```

PR #84, opened 2019-01-16
Editorial: Explicit sequence number

emanjon:
The draft states in three places that DTLS use explicit sequence numbers, while this was true for DTLS 1.2, the sequence number in the DTLSCipherstructure is a mixture of implicit and explict. Maybe "Partly Explicit"?


ekr:
Please provide me with a PR to add your name to acknowledgements.

emanjon:
Hi,

I don’t know what PR is an abbreviation for but I guess it is not clear from my username who I am. I should probably update it, I did not intend to be anonymous :)

John Mattsson, Ericsson, john.mattsson@ericsson.com

PR #85, opened 2019-04-02
Various editorial fixes.

davidben:
I read through the draft and amassed a number of comments. This PR resolves the easy stuff. I'll send an email to the WG list with the more interesting ones shortly.

- Grammar fixes.

- The text describing sequence number encryption doesn't say how the
  receiver handles short ciphertexts. It's also phrased oddly. It says
  "In some (rare) cases the ciphertext may be less than 16 bytes", but
  the whole point of the text is to ensure it cannot be less than 16
  bytes. Reword it.

- Whether there's a Certificate message is no longer a function of the
  ciphersuite, but of parameters in general. (With resumption that was
  already true.)

- The bits talking about epoch 1-3 so far before the definitions of
  those epochs should link to the section.

- Remove some remnants of EndOfEarlyData. They're still in the
  definition of the Handshake struct. I wasn't sure how to handle that
  since it was in the HandshakeType enum and we probably shouldn't call
  it RESERVED when it exists in TLS?

- Spell out [sender]\_application\_traffic\_secret\_\* since traffic\_secret\_\*
  isn't defined.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -541,7 +541,7 @@ MUST reject shorter records as if they had failed deprotection, as described in
 conventional record padding mechanism) in order to make a suitable-length
 ciphertext. Note most of the DTLS AEAD algorithms have a 16-byte authentication
 tag and need no padding. However, some algorithms such as
-TLS_AES_128_CCM_8_SHA256 have a shorter authentication tag and require padding
+TLS_AES_128_CCM_8_SHA256 have a shorter authentication tag and may require padding
 for short inputs.
 
 Note that sequence number encryption is only applied to the DTLSCiphertext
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -535,14 +535,14 @@ The encrypted sequence number is computed by XORing the leading
...
-less than 16 bytes. In this case, implementations MUST pad the
-plaintext out (using the conventional record padding mechanism)
-in order to make a suitable-length ciphertext.
+This procedure requires the ciphertext length be at least 16 bytes. Receivers
+MUST reject shorter records as if they had failed deprotection, as described in
+{{handling-invalid-records}}. Senders MUST pad short plaintexts out (using the
+conventional record padding mechanism) in order to make a suitable-length
+ciphertext. Note most of the DTLS AEAD algorithms have a 16-byte authentication
+tag and need no padding. However, some algorithms such as
+TLS_AES_128_CCM_8_SHA256 have a shorter authentication tag and require padding
```
```suggestion
TLS_AES_128_CCM_8_SHA256 have a shorter authentication tag and may require padding
```

hannestschofenig:
This looks good to me.

draft-ietf-tls-dtls13.md
hannestschofenig: I fixed this issue

PR #87, opened 2019-04-22
Note about additional data handling for record layer protection

hannestschofenig:


hannestschofenig:
I added this remark following my implementation experience of DTLS 1.2 CID followed by the work on the DTLS 1.3. We noticed this issue at the Prague IETF meeting (March 2019).

PR #89, opened 2019-04-30
A right for the server to reboot

ximaera:
IMO it's just an editorial suggestion to make a use case more clear to a reader.


hannestschofenig:
This extra text introduces possibilities for DoS attacks. The case where the server does not have enough memory to delay the deletion of the old association before completing the return routability check does not appear very common to me.

ximaera:
May I ask you to elaborate a bit on the DoS scenario you've mentioned?

Resource constrained environment is typical for IoT devices which are among the most prominent DTLS users.

ekr:
I agree with Hannes that this doesn't actually help. If you're really low on memory you can just abandon associations whenever.

Issue #77, opened 2018-11-19
Who restarts the connection after receiving a bad cookie?


martinthomson:
> If a server receives a ClientHello with an invalid cookie, it MUST NOT respond with a HelloRetryRequest. Restarting the handshake from scratch, without a cookie, allows the client to recover from a situation where it obtained a cookie that cannot be verified by the server.

I don't understand what this is trying to say.  I think that it says that the client, having received a bad cookie will send its second ClientHello off into the void, only to have the handshake time out.  Then, the client is expected to restart.

This is probably right.  The bad cookie indicates that something is badly wrong.  Anything the server might send that would fix this wouldn't be authenticated and so might be exploited by an attack as a denial of service on the connection.  However, I'm not sure if any client actually implements this restart.

ekr:
Probably not, but I think this is just editorial. If you had some client-server type app then you might expect a retry (like Web clients do). Anyway, we shouldn't change the spec.

hannestschofenig:
The text in DTLS 1.2 was a bit different, namely 
" 
   If a server receives a ClientHello with an invalid cookie, it SHOULD
   treat it the same as a ClientHello with no cookie.  This avoids
   race/deadlock conditions if the client somehow gets a bad cookie
   (e.g., because the server changes its cookie signing key).
"

I believe the DTLS 1.2 text was actually quite good. I suggest to re-use the DTLS 1.2 text again. 
We also have text in TLS 1.3 to deal with the case where the Client receives a second HelloRetryRequest:  

"
   If a client receives a second
   HelloRetryRequest in the same connection (i.e., where the ClientHello
   was itself in response to a HelloRetryRequest), it MUST abort the
   handshake with an "unexpected_message" alert.
"

martinthomson:
I don't believe that the bad cookie is a situation that needs to be handled in a different way than is described.  Any server doing this uses a key identifier in its cookie.  So, like ekr, I don't think that the design needs to be changed.

I would prefer to say that the ClientHello with a bad cookie is simply discarded (as Internet background noise, if you will).  If the intent of the text (prior to #91 as proposed) is to say that it's now the client's problem to fix, I agree with that.  The question was about what the text meant, so an editorial fix would be fine.  I would go so far as to say:

> If a server receives a ClientHello with an invalid cookie, it MUST discard/ignore the record.

PR #91, opened 2019-07-07
Dealing with invalid cookies

hannestschofenig:
as raised by Martin in https://github.com/tlswg/dtls13-spec/issues/77


ekr:
This is not what we agreed, so closing this.

PR #97, opened 2019-07-08
Davidben comments

ekr:


ekr:
@davidben @martinthomson

draft-ietf-tls-dtls13.md
```diff
@@ -681,7 +694,9 @@ receiver MUST verify that the record contains a sequence number that
 does not duplicate the sequence number of any other record received
 during the lifetime of the session. This check SHOULD happen after
 deprotecting the packet; otherwise the packet discard might itself
-serve as a timing channel for the sequence number.
+serve as a timing channel for the sequence number. Note that decompressing
+the record number is still a potential timing channel, though a weaker
+one.
```
martinthomson: Weaker seems - for want of a better word - weak.  The consequence here is that you expose some information about the sequence number. That doesn't provide an attacker much advantage in terms of assisting in key recovery because the key is independent of other keys and the protected information is very low entropy anyway.

draft-ietf-tls-dtls13.md
```diff
@@ -412,6 +412,19 @@ protected with either \[sender]_handshake_traffic_sercret or
...
+unpacked RecordNumber structure, as shown below:
+
+~~~
+    struct {
+        uint16 epoch;
+        uint48 sequence_number;
+    } RecordNumber;
+~~~
+
+This value is used in the ACK message as well as in the "record_sequence_number"
```
```suggestion
This 64-bit value is used in the ACK message as well as in the "record_sequence_number"
```

ekr:
Well, clients can recover by tearing down the connection and starting over.

chris-wood:
> Well, clients can recover by tearing down the connection and starting over.

Sure, sure. I think this was in the context of a "single" connection. Perhaps it's not that important to clarify?

ekr:
I'll add text.

PR #99, opened 2019-10-01
Change the key label for DTLS to be "dtls13". Note that we have to

ekr:
omit the trailing space because TLS 1.3 went right up to the limit.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1353,7 +1353,7 @@ the algorithm described in Section 4.4.3 and Section 4.4.4 of {{!TLS13}}, respec
 ##  Cryptographic Label Prefix
 
 Section 7.1 of {{RFC8446}} specifies that HKDR-Expand-Label uses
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1350,6 +1350,15 @@ if each handshake message had been sent as a single fragment following
 the algorithm described in Section 4.4.3 and Section 4.4.4 of {{!TLS13}}, respectively.
 
 
+##  Cryptographic Label Prefix
+
+Section 7.1 of {{RFC8446}} specifies that HKDR-Expand-Label uses
```
davidben: Nit: HKDF-Expand-Label (not HKDR).

ekr:
No, no, DTLS also uses HKDR for Key Separation++

draft-ietf-tls-dtls13.md
```suggestion
Section 7.1 of {{RFC8446}} specifies that HKDF-Expand-Label uses
```

chris-wood:
Hearing no objections, we're merging this change. I'll fix the HKDR->HKDF typo in a followup PR.

PR #103, opened 2019-11-04
adding DTLS-OK column values

### LINKED ISSUE #102: IANA Considerations
content types and handshake types needs to say whether they are DTLS-OKAY.  The ones defined in the DTLS spec obviously are, but to ensure IANA gets clear instructions we should include a value for these columns.

seanturner:
Closes #102.


ekr:
merged by hand

PR #111, opened 2019-11-09
Issue110 multiple connections same dg

ekr:


ekr:
@martinthomson

draft-ietf-tls-dtls13.md
```diff
@@ -404,13 +404,17 @@ The length field MAY be omitted by clearing the L bit, which means that the
...
+Omitting the length field MUST only be used for the last record in a
+datagram.
+
+Implementations which send multiple records in the same datagram
+SHOULD omit the connection id from all but the first record; receiving
+implementations MUST assume that any subsequent records without
+connection IDs belong to the same assocatiation.  Sending
+implementations MUST NOT mix records from multiple DTLS associations
+in the same datagram. If the second or later record has a connection
+ID which does not match that used for previous records, the rest of
```
martinthomson: "That does not correspond to the same connection"?  Do we want to allow for change of connection ID for the same connection?  It allows for correlation between the two, so it's hardly a good idea, but I don't see why it couldn't be done.

ekr:
Merged by hand after some git screwups

PR #112, opened 2019-11-09
Issue109 chacha order

ekr:


ekr:
@davidben @martinthomson

PR #129, opened 2020-03-29
Removed term "deprotection"

hannestschofenig:


ekr:
I don't see a reason for this change 8446 already uses deprotect.

hannestschofenig:
The problem is that the TLS 1.3 spec does not define the term either.

ekr:
Well, because I thought it was obvious. However, if you can't understand TLS 1.3, then you can't understand DTLS 1.3, so I think the ship has sailed.

ekr:
@hannestschofenig I think this is within our discretion and I think we should close it. LMK if you feel strongly.

PR #130, opened 2020-03-29
Update to Figure 1

hannestschofenig:


draft-ietf-tls-dtls13.md
```diff
@@ -204,7 +204,7 @@ phase of the DTLS handshake:
          ------                                   ------
          ClientHello           ------>
 
-                                 X<-- HelloRetryRequest
+                                 X<----       ServerHello
```
ekr, line 207: You need to delete a space on the left here.

ekr:
Fixed by hand

PR #139, opened 2020-04-02
Clarify when to ACK and what to ACK

hanno-arm:
In response to feedback on the mailing list (https://mailarchive.ietf.org/arch/msg/tls/LMn0qma6t896esRO44DwaS_54VM/), this PR attempts to clarify (a) when to generate ACKs, and (b) which records to ACK.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1616,11 +1616,12 @@ simply use the current key.
 
 ## Sending ACKs
 
-When an implementation receives a partial flight, it SHOULD generate
-an ACK that covers the messages from that flight which it has
-received so far. Implementations have some discretion about when
-to generate ACKs, but it is RECOMMENDED that they do so under
-two circumstances:
+When an implementation detects a disruption in the receipt of the
+current incoming flight, it SHOULD generate an ACK that covers the
+messages from that flight which it has received and processed so far.
+Implementations have some discretion about which events to treat
+as signs of disruption, but it is RECOMMENDED that they generate
+ACKs under two circumstances:
 
 - When they receive a message or fragment which is out of order,
   either because it is not the next expected message or because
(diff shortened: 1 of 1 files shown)


hanno-arm:
@jfhamme-cccs Thank you for your feedback! I have removed the sentence because its content is already implied by what's stated in the previous section [ACK message](https://tools.ietf.org/html/draft-ietf-tls-dtls13-37#section-7). See the commit description of https://github.com/tlswg/dtls13-spec/pull/139/commits/f22743cb31e444b4efeabbbca5a69d54a0b3f8a8.

thomas-fossati:
@hanno-arm thanks.  This sensibly improves readability and clarity.

draft-ietf-tls-dtls13.md
```diff
@@ -1594,8 +1594,10 @@ record_numbers:
   order.
 
 Implementations MUST NOT acknowledge records containing
-non-duplicative handshake messages or fragments which have not been
+handshake messages or fragments which have not been
 processed or buffered. Otherwise, deadlock can ensue.
+As an example, implementations MUST NOT send ACKs for
+handshake messages which they discard as out-of-order.
```
ekr, line 1600: This is actually different. Consider the case where I receive:

```
1: MSG=1 [1:100]
2: MSG=2 [1:100]
3: MSG=1 [1:50]
```
I may discard three because it's duplicative and/or out of order, but I should ACK it.

hanno-arm: Ah, that's what's meant here - @thomas-fossati and I were pondering for a bit, but didn't get the intended meaning of non-duplicative. I agree that in your example, msg 3 should be ACKed.

I think the issue here is the ambiguity of what 'processed' means - I'd say that message 3 will still be 'processed', even though it will be noted that its content doesn't reveal anything new.

Maybe we can find a wording which is less prone to different interpretations.

PR #142, opened 2020-04-19
Ack epochs2

ekr:
Clarify the sending epoch.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1616,8 +1616,8 @@ it must do so in epoch 2, as it is required to use an epoch
 greater than or equal to 2 and cannot yet send with any greater
 epoch. Implementations SHOULD simply use the highest 
 current sending epoch, which will generally be the highest available.
-After the handshake, implementations MUST use the highest available
-sending epoch.
+After the handshake, implementations MUST use the current sending epoch
+for sending ACKs.
 
 
 ## Sending ACKs
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1610,7 +1610,7 @@ cover more than one flight.
 During the handshake, ACK records MUST be sent with an epoch that is
 equal to or higher than the record which is being acknowledged.
 Note that some care is required when processing flights spanning
-multiple epochs. For instance, if the client receives only the Server Hello
+multiple epochs. For instance, if the client receives only the ServerHello
 and Certificate and wishes to ACK them in a single record, 
 it must do so in epoch 2, as it is required to use an epoch
 greater than or equal to 2 and cannot yet send with any greater
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1616,7 +1616,8 @@ it must do so in epoch 2, as it is required to use an epoch
 greater than or equal to 2 and cannot yet send with any greater
 epoch. Implementations SHOULD simply use the highest 
 current sending epoch, which will generally be the highest available.
-
+After the handshake, implementations MUST use the highest available
+sending epoch.
 
 
 ## Sending ACKs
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1607,9 +1607,17 @@ After the handshake, ACKs SHOULD be sent once for each received
 and processed handshake record (potentially subject to some delay) and MAY
 cover more than one flight.
 
-ACK records MUST be sent with an epoch that is equal to or higher
-than the record which is being acknowledged. Implementations SHOULD
-simply use the current key.
+During the handshake, ACK records MUST be sent with an epoch that is
+equal to or higher than the record which is being acknowledged.
+Note that some care is required when processing flights spanning
+multiple epochs. For instance, if the client receives only the Server Hello
```
hanno-arm, line 1613: Trivial: 
```suggestion
multiple epochs. For instance, if the client receives only the ServerHello
```

draft-ietf-tls-dtls13.md
```diff
@@ -1607,9 +1607,17 @@ After the handshake, ACKs SHOULD be sent once for each received
 and processed handshake record (potentially subject to some delay) and MAY
 cover more than one flight.
 
-ACK records MUST be sent with an epoch that is equal to or higher
-than the record which is being acknowledged. Implementations SHOULD
-simply use the current key.
+During the handshake, ACK records MUST be sent with an epoch that is
+equal to or higher than the record which is being acknowledged.
```
hanno-arm, line 1611: This is a bit pedantic, but strictly speaking this would allow a Client to ACK a ServerHello with epoch 1 (early data) - this doesn't seem right, because the server might not accept it.

draft-ietf-tls-dtls13.md
```diff
@@ -1607,9 +1607,17 @@ After the handshake, ACKs SHOULD be sent once for each received
...
+Note that some care is required when processing flights spanning
+multiple epochs. For instance, if the client receives only the Server Hello
+and Certificate and wishes to ACK them in a single record, 
+it must do so in epoch 2, as it is required to use an epoch
+greater than or equal to 2 and cannot yet send with any greater
+epoch. Implementations SHOULD simply use the highest 
+current sending epoch, which will generally be the highest available.
+After the handshake, implementations MUST use the highest available
+sending epoch.
```
hanno-arm, line 1620: I find this slightly confusing: After the handshake, as far as I understand there's an unambiguously defined 'current outgoing epoch', and that's what should always be used, including ACKs. Talking of the 'highest available epoch' suggests there is some leeway which I don't see at the moment. 
In particular, even though sending a `KeyUpdate` suggests the transition to the next sending epoch which could hence be seen 'available', it must not be used before the `KeyUpdate` has been ACKed.

Can't we simply say:
```suggestion
After the handshake, implementations MUST use the current sending epoch
for sending ACKs.
```

draft-ietf-tls-dtls13.md
ekr: Sorry, I had QUIC epochs in my head. Will fix.

draft-ietf-tls-dtls13.md
ekr: Right. I'll change.

draft-ietf-tls-dtls13.md
ekr, line 1611:
```suggestion
equal to or higher than the record which is being acknowledged (with the exception of epoch 1, which MUST NOT be used for sending ACKs).
```

chris-wood:
This LGTM, pending merge conflicts.

draft-ietf-tls-dtls13.md
```diff
@@ -1607,9 +1607,17 @@ After the handshake, ACKs SHOULD be sent once for each received
 and processed handshake record (potentially subject to some delay) and MAY
 cover more than one flight.
 
-ACK records MUST be sent with an epoch that is equal to or higher
-than the record which is being acknowledged. Implementations SHOULD
-simply use the current key.
+During the handshake, ACK records MUST be sent with an epoch that is
+equal to or higher than the record which is being acknowledged (with the exception of epoch 1, which MUST NOT be used for sending ACKs).
```
hanno-arm, line 1611: Trivial style change:
```suggestion
equal to or higher than the record which is being acknowledged (with 
the exception of epoch 1, which MUST NOT be used for sending ACKs).
```

ekr:
Merged by hand

PR #147, opened 2020-05-13
Define anti-forgery limits for DTLS

### LINKED ISSUE #145: Integrity bounds
We should count failed "deprotection" attempts and limit the number that are permitted with any given key.  Otherwise, an attacker might be able to successfully forge a record.

See https://github.com/quicwg/base-drafts/issues/3619 for a more thorough description of the problem and the analysis of different options.

martinthomson:
And define limits for the CCM ciphers.

This turned out to be a little tricky.  I copied the text about
AEAD_AES_128_CCM from the QUIC PR, which is easy, but AEAD_AES_128_CCM_8
is a while different story.  As that is defined in TLS 1.3, it seems
necessary to address it here.  QUIC simply doesn't allow that, but I
assume that DTLS will.

On the basis that DTLS will allow AEAD_AES_128_CCM_8, we need some
language about it.  I've tried to write something, but I'm not very
happy with it.  I'd greatly prefer not to allow this AEAD in DTLS.  It's
basically fine in TLS, but here the margins are just too narrow.  

I realize that 2^-57 is essentially arbitrary, you might reasonably say that 2^-40 is better, but that would change the analysis in other ways.

Closes #145.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -63,6 +63,21 @@ informative:
   RFC5246:
   RFC6347:
   RFC7525:
+  AEBounds:
+    title: "Limits on Authenticated Encryption Use in TLS"
+    author:
+      - ins: A. Luykx
+      - ins: K. Paterson
+    date: 2016-03-08
+    target: "http://www.isg.rhul.ac.uk/~kp/TLS-AEbounds.pdf"
+  ROBUST:
+    title: "Robust Channels: Handling Unreliable Networks in the Record Layers of QUIC and DTLS"
+    author:
+      - ins: M. Fischlin
+      - ins: F. Günther
+      - ins: C. Janson
+    date: 2020-02-21
+    target: "https://www.felixguenther.info/docs/QUIPS2020_RobustChannels.pdf"
 
 --- abstract
 
@@ -748,6 +763,55 @@ If DTLS is being carried over a transport that is resistant to
 forgery (e.g., SCTP with SCTP-AUTH), then it is safer to send alerts
 because an attacker will have difficulty forging a datagram that will
 not be rejected by the transport layer.
+
+
+### AEAD Limits
+
+Section 5.5 of TLS {{!TLS13}} defines limits on the number of records that can
+be protected using the same keys. These limits are specific to an AEAD
+algorithm, and apply equally to DTLS. Implementations SHOULD NOT protect more
+records than allowed by the limit specified for the negotiated AEAD.
+
+{{!TLS13}} does not specify a limit for AEAD_AES_128_CCM, but the analysis in
+{{ccm-bounds}} shows that a limit of 2^23 packets can be used to obtain the
+same confidentiality protection as the limits specified in TLS.
+
+The usage limits defined in TLS 1.3 exist for protection against attacks
+on confidentiality and apply to successful applications of AEAD protection. The
+integrity protections in authenticated encryption also depend on limiting the
+number of attempts to forge packets. TLS achieves this by closing connections
+after any record fails an authentication check. In comparison, DTLS ignores any
+packet that cannot be authenticated, allowing multiple forgery attempts.
+
+Implementations MUST count the number of received packets that fail
+authentication. If the number of packets that fail authentication exceeds a
+limit that is specific to the AEAD in use, an implementation MUST immediately
+close the connection. Implementations SHOULD initiate a key update before
+reaching this limit. Applying a limit reduces the probability that an attacker
+is able to successfully forge a packet; see {{AEBounds}} and {{ROBUST}}.
+
+For AEAD_AES_128_GCM, AEAD_AES_256_GCM, and AEAD_CHACHA20_POLY1305, the limit
+on the number of records that fail authentication is 2^36. Note that the
+analysis in {{AEBounds}} supports a higher limit for the AEAD_AES_128_GCM and
+AEAD_AES_256_GCM, but this specification recommends a lower limit. For
+AEAD_AES_128_CCM, the limit on the number of records that fail authentication
+is 2^23.5; see {{ccm-bounds}}.
+
+The AEAD_AES_128_CCM_8 AEAD, as used in TLS_AES_128_CCM_SHA256, does not have a
+limit on the number of records that fail authentication that both limits the
+probability of forgery by the same amount and does not expose implementations
+to the risk of denial of service; see {{ccm-short}}. Therefore,
+TLS_AES_128_CCM_SHA256 MUST NOT used in DTLS without additional safeguards
+against forgery. Implementations MUST set usage limits for AEAD_AES_128_CCM_8
+based on an understanding of any additional forgery protections that are used.
+
+Any TLS cipher suite that is specified for use with DTLS MUST define limits on
+the use of the associated AEAD function that preserves margins for both
+confidentiality and integrity. That is, limits MUST be specified for the number
+of packets that can be authenticated and for the number packets that can fail
+authentication. Providing a reference to any analysis upon which values are
+based - and any assumptions used in that analysis - allows limits to be adapted
+to varying usage conditions.
 
 
 # The DTLS Handshake Protocol {#dtls}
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,55 @@ because an attacker will have difficulty forging a datagram that will
...
+number of attempts to forge packets. TLS achieves this by closing connections
+after any record fails an authentication check. In comparison, DTLS ignores any
+packet that cannot be authenticated, allowing multiple forgery attempts.
+
+Implementations MUST count the number of received packets that fail
+authentication. If the number of packets that fail authentication exceeds a
+limit that is specific to the AEAD in use, an implementation MUST immediately
+close the connection. Implementations SHOULD initiate a key update before
+reaching this limit. Applying a limit reduces the probability that an attacker
+is able to successfully forge a packet; see {{AEBounds}} and {{ROBUST}}.
```
ekr, line 791: A few points:

1. You need to initiate a key update with UpdateRequested, right?
2. The limit is actually per key, right?
3. Once the update has started, you could presumably drop the new epoch keys early rather than terminating the connection.

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,55 @@ because an attacker will have difficulty forging a datagram that will
...
+
+For AEAD_AES_128_GCM, AEAD_AES_256_GCM, and AEAD_CHACHA20_POLY1305, the limit
+on the number of records that fail authentication is 2^36. Note that the
+analysis in {{AEBounds}} supports a higher limit for the AEAD_AES_128_GCM and
+AEAD_AES_256_GCM, but this specification recommends a lower limit. For
```
ekr, line 796: Why does it recommend that?

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,55 @@ because an attacker will have difficulty forging a datagram that will
...
+AEAD_AES_128_CCM, the limit on the number of records that fail authentication
+is 2^23.5; see {{ccm-bounds}}.
+
+The AEAD_AES_128_CCM_8 AEAD, as used in TLS_AES_128_CCM_SHA256, does not have a
+limit on the number of records that fail authentication that both limits the
+probability of forgery by the same amount and does not expose implementations
+to the risk of denial of service; see {{ccm-short}}. Therefore,
+TLS_AES_128_CCM_SHA256 MUST NOT used in DTLS without additional safeguards
+against forgery. Implementations MUST set usage limits for AEAD_AES_128_CCM_8
+based on an understanding of any additional forgery protections that are used.
```
ekr, line 806: Do we have some idea what those protections might be?

draft-ietf-tls-dtls13.md
```diff
@@ -2016,6 +2080,132 @@ This section provides the normative protocol types and constants definitions.
 %%## ACKs
 %%## Connection ID Management
 
+# Analysis of Limits on CCM Usage {#ccm-bounds}
```
ekr, line 2083: I haven't read this section yet, so this is a TODO for me.

draft-ietf-tls-dtls13.md
martinthomson: Because we felt like it.

We could split the difference and use 2^60 for GCM, but it's a ludicrous number.

draft-ietf-tls-dtls13.md
martinthomson: My only thought here is that you might have a hard limit on the number of packets you can receive, ever that makes larger limits meaningless.

More realistically, this probably comes down to having a higher tolerance for forgeries in that application.  In the real-time case, the window over which a forgery has any effect is pretty narrow.  For instance, new video I-frames replace the entire scene.  So I can see how someone might justify a higher tolerance for forgery.

draft-ietf-tls-dtls13.md
martinthomson: 1. Good point.  QUIC doesn't have this distinction, but it matters here.

2. Yes, per key.  Will clarify (and in QUIC too).

3. That's a good point.  I'll add something.

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,58 @@ because an attacker will have difficulty forging a datagram that will
 not be rejected by the transport layer.
 
 
+### AEAD Limits
+
+Section 5.5 of TLS {{!TLS13}} defines limits on the number of records that can
+be protected using the same keys. These limits are specific to an AEAD
+algorithm, and apply equally to DTLS. Implementations SHOULD NOT protect more
+records than allowed by the limit specified for the negotiated AEAD.
```
hanno-arm, line 773: One might mention here that the recommended action is to issue a `KeyUpdate` to update the keys (in particular, it's not necessary to terminate the connection).

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,58 @@ because an attacker will have difficulty forging a datagram that will
 not be rejected by the transport layer.
 
 
+### AEAD Limits
+
+Section 5.5 of TLS {{!TLS13}} defines limits on the number of records that can
+be protected using the same keys. These limits are specific to an AEAD
+algorithm, and apply equally to DTLS. Implementations SHOULD NOT protect more
```
hanno-arm, line 772: Why is this a SHOULD NOT and not a MUST NOT?

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,58 @@ because an attacker will have difficulty forging a datagram that will
...
+
+{{!TLS13}} does not specify a limit for AEAD_AES_128_CCM, but the analysis in
+{{ccm-bounds}} shows that a limit of 2^23 packets can be used to obtain the
+same confidentiality protection as the limits specified in TLS.
+
+The usage limits defined in TLS 1.3 exist for protection against attacks
+on confidentiality and apply to successful applications of AEAD protection. The
+integrity protections in authenticated encryption also depend on limiting the
+number of attempts to forge packets. TLS achieves this by closing connections
+after any record fails an authentication check. In comparison, DTLS ignores any
```
hanno-arm, line 783: IIRC, the spec recommends to do so, but it doesn't enforce it.

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,58 @@ because an attacker will have difficulty forging a datagram that will
...
+packet that cannot be authenticated, allowing multiple forgery attempts.
+
+Implementations MUST count the number of received packets that fail
+authentication. If the number of packets that fail authentication with a single
```
hanno-arm, line 787: Nitpick: As detailed in the next sentence, what needs to be counted is the number of failed authentications for a particular key, not the global number.

draft-ietf-tls-dtls13.md
```diff
@@ -750,6 +765,58 @@ because an attacker will have difficulty forging a datagram that will
...
+key exceeds a limit that is specific to the AEAD in use, an implementation MUST
+immediately close the connection. Implementations SHOULD initiate a key update
+with update_requested before reaching this limit. Once a key update has been
+initiated, the previous keys can be dropped when the limit is reached rather
+than closing the connection. Applying a limit reduces the probability that an
+attacker is able to successfully forge a packet; see {{AEBounds}} and
+{{ROBUST}}.
+
+For AEAD_AES_128_GCM, AEAD_AES_256_GCM, and AEAD_CHACHA20_POLY1305, the limit
+on the number of records that fail authentication is 2^36. Note that the
```
hanno-arm, line 797: "... the number of records ... _is_ 2^36." is quite strong wording, and in particular it's not clear where this number comes from and whether it's taken from one of the papers, a universal, or our own recommendation. I'd suggestt to either name the source here, or add a half sentence like "this specification recommends ..."

draft-ietf-tls-dtls13.md
martinthomson: Because TLS 1.3 uses SHOULD NOT.

draft-ietf-tls-dtls13.md
martinthomson: >    If the decryption fails, the receiver MUST terminate the connection with a "bad_record_mac" alert.
-- https://tools.ietf.org/html/rfc8446#section-5.2

That seems pretty clear to me.

draft-ietf-tls-dtls13.md
martinthomson: Willdo.

draft-ietf-tls-dtls13.md
martinthomson: Sounds good.

draft-ietf-tls-dtls13.md
martinthomson: I'm using a SHOULD instead of MUST for the limit now, so I don't feel like the definitive statement needs to be softened any.

As far as it goes, if you feel like you can drop >60 billion bogus records and keep going, then you are welcome to do so.

draft-ietf-tls-dtls13.md
```diff
@@ -2016,6 +2083,136 @@ This section provides the normative protocol types and constants definitions.
 %%## ACKs
 %%## Connection ID Management
 
+# Analysis of Limits on CCM Usage {#ccm-bounds}
```
hanno-arm, line 2087: General question: Is this analysis something that should be part of the spec, or should it rather be published separately and only referenced here?

draft-ietf-tls-dtls13.md
hanno-arm: No, I mean the DTLS part.

draft-ietf-tls-dtls13.md
```diff
@@ -2016,6 +2084,136 @@ This section provides the normative protocol types and constants definitions.
...
+
+TLS {{?TLS13}} and {{AEBounds}} do not specify limits on usage for
```
hanno-arm, line 2089: Nitpick: `usage` -> `key usage`.

draft-ietf-tls-dtls13.md
```diff
@@ -2016,6 +2084,136 @@ This section provides the normative protocol types and constants definitions.
...
+
+
+## Confidentiality Limits {#ccm-confidentiality}
+
+For confidentiality, Theorem 2 in {{?CCM-ANALYSIS}} establishes that an attacker
+gains a distinguishing advantage over an ideal pseudorandom permutation (PRP) of
+no more than:
+
+~~~
+(2l * q)^2 / 2^n
```
hanno-arm, line 2146: General remark on the analysis: While its result is stated in terms of number of packets, assuming a fixed (large) size per packet, the analysis itself seems to rather involve the total number of blocks that are being protected. For some applications with very small packet sizes, the derived recommendations in terms of packet numbers will therefore unnecessarily strict. Given that the derived constraints are of practical relevance, I wonder if it's worth stating the analysis in terms of total number of blocks, and derive a less precise but safe bound in terms of the number of packets from that.

martinthomson: This is something we discussed with respect to TLS as well.  Though you might get larger limits if you count blocks, doing so is not natural in (D)TLS.  If you think that a key update every 2^20-something is onerous, then you might want to count more closely, but the large number means that you don't really need to quibble about whether it is 2^24.5 or 2^29.5.

hanno-arm: Does it make a noteworthy difference for CCM-8, though, where the proposed integrity bound assuming full-length packets is quite low?

hannestschofenig:
We cannot get rid of CCM_8 because this is used in various IoT standards and it is also implemented in hardware.

martinthomson:
> We cannot get rid of CCM_8

That's why this PR is shaped as it is.  Some of us which we could get rid of it though; it's just not strong enough.

PR #149, opened 2020-05-17
Clear up that the ACK logic is the union

ekr:


chris-wood:
This is editorial, so I'm merging as is.

PR #164, opened 2020-12-06
Clean up use of connection and association

ekr:
This should make things correct. We might consider getting rid of the term connection or association entirely, but each is problematic: association is a historical term but we also use connection a lot anyway.


martinthomson (approved):
It's a shame that "association" is so firmly entrenched.  I don't see "association" being used much anyway.  Good that you have removed "session" though.

chris-wood (approved):

PR #166, opened 2020-12-06
Require that cipher suites define a record number encryption algorithm

ekr:


martinthomson (approved):

draft-ietf-tls-dtls13.md
```diff
@@ -2088,6 +2092,9 @@ registry, defined in {{!TLS13}}, for RequestConnectionId (TBD), and
 NewConnectionId (TBD), as defined in this document.  The value for the
 "DTLS-OK" columns are "Y".
 
+IANA [SHALL mark/has marked] the cipher suites TLS_SHA256_SHA256 and
+TLS_SHA384_SHA_384 DTLS=N.
```
chris-wood, line 2096:
```suggestion
IANA [SHALL set] the "DTLS-OK" column value for cipher suites TLS_SHA256_SHA256 and
TLS_SHA384_SHA_384 to "N".
```

chris-wood (approved):
This one requires WG consensus, I think.

draft-ietf-tls-dtls13.md
```diff
@@ -582,6 +582,10 @@ tag and need no padding. However, some algorithms such as
 TLS_AES_128_CCM_8_SHA256 have a shorter authentication tag and may require padding
 for short inputs.
 
+Future cipher suites which are not based on AES or ChaCha20 MUST define
+their own record sequence number encryption in order to be used with
+DTLS.
```
kaduk, line 587: I can't decide whether this requirement merits adding this document as a reference for the ciphersuites registry.

martinthomson: I would say yes, provided that it is additive.

chris-wood: Agreed.

hannestschofenig: What does this imply for the document itself? Do we need to capture it in the IANA consideration section or would we ask IANA to update the references for the TLS Cipher Suites registry at https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml?

draft-ietf-tls-dtls13.md
hannestschofenig: Why do we need to add this IANA entry in this document? Shouldn't https://tools.ietf.org/html/draft-camwinget-tls-ts13-macciphersuites-08 add it to IANA given that they are defining the ciphersuite?

draft-ietf-tls-dtls13.md
kaduk: Typically we would have this document's IANA Considerations section request IANA to update the references for the registry to add this document as a reference.

draft-ietf-tls-dtls13.md
kaduk: I agree with Hannes; I don't think we need to do anything about the MAC-only ciphersuites entries.
I've already been in contact with the authors (and DEs) and the authors plan to add text to handle this case.  (Though, they claim that the record numbers will just not be encrypted and that's okay.)

ekr: They are already in the registry and marked DTLS-OK.

draft-ietf-tls-dtls13.md
ekr: Done.

draft-ietf-tls-dtls13.md
kaduk: Yes, they are.  Is that wrong?  Is that more wrong than having MAC-only ciphersuites in the first place?

ekr:
I guess not. Ugh.

draft-ietf-tls-dtls13.md
```diff
@@ -2081,13 +2085,17 @@ IANA is requested to allocate a new value in the "TLS ContentType"
 registry for the ACK message, defined in {{ack-msg}}, with content type 26.
 The value for the "DTLS-OK" column is "Y".  IANA is requested to reserve
 the content type range 32-63 so that content types in this range are not
-allocated.
+allocated.
```
chris-wood, line 2088:
```suggestion
allocated.
```

draft-ietf-tls-dtls13.md
ekr: Yeah, I guess we have to live with this.

draft-ietf-tls-dtls13.md
```diff
@@ -2194,6 +2198,10 @@ registry, defined in {{!TLS13}}, for RequestConnectionId (TBD), and
 NewConnectionId (TBD), as defined in this document.  The value for the
 "DTLS-OK" columns are "Y".
 
+IANA is requested to add this RFC as a reference to the TLS Cipher Suite Registry.
+IANA [SHALL set] the "DTLS-OK" column value for cipher suites TLS_SHA256_SHA256 and
+TLS_SHA384_SHA_384 to "N".
+
```
chris-wood, line 2204:
```suggestion
```

PR #168, opened 2020-12-06
PMTU estimates

ekr:


draft-ietf-tls-dtls13.md
```diff
@@ -654,16 +654,19 @@ protocols. In particular:
...
 implementations SHOULD ignore any such messages that indicate
 PMTUs below the IPv4 and IPv6 minimums of 576 and 1280 bytes
 respectively
 
-The DTLS record layer SHOULD allow the upper layer protocol to
-discover the amount of record expansion expected by the DTLS
-processing.
-
-If there is a transport protocol indication (either via ICMP or via a
+If there is a transport protocol indication of a exceeding the PMTU
```
martinthomson, line 668:
```suggestion
If there is a transport protocol indication that the PMTU was exceeded
```

martinthomson (approved):

chris-wood (approved):
This one probably needs consensus given new normative language.

hannestschofenig:
I am OK with this change. There are no normative changes in there because the normative text segment is only moved to a different paragraph.

chris-wood:
I was referring to this added clause: "alternately it *MAY* report PMTU estimates minus the estimated expansion from the transport layer and DTLS record framing."

martinthomson:
What is awkward here is that this is phrased as normative requirements on an API.  Maybe worded as advice "an application that uses DTLS might need information from DTLS to allow it to estimate the PMTU; DTLS could report a PMTU estimate less any DTLS overheads, or it can report those overheads and allow the application to perform its own estimation"

ekr:
There has historically been API language here, so I think we should keep it and just adjust it as here.

PR #169, opened 2020-12-06
Updates to cookie text

ekr:


draft-ietf-tls-dtls13.md
```diff
@@ -866,8 +866,9 @@ cookie technique used by Photuris {{RFC2522}} and IKE {{RFC7296}}.  When
 the client sends its ClientHello message to the server, the server
 MAY respond with a HelloRetryRequest message. The HelloRetryRequest message,
 as well as the cookie extension, is defined in TLS 1.3.
-The HelloRetryRequest message contains a stateless cookie generated using
-the technique of {{RFC2522}}. The client MUST retransmit the ClientHello
+The HelloRetryRequest message contains a stateless cookie (see
+{{RFC8446}}; Section 4.2.2.).
```
martinthomson, line 870:
```suggestion
{{RFC8446}}; Section 4.2.2).
```

draft-ietf-tls-dtls13.md
```diff
@@ -956,6 +957,10 @@ configured not to perform a cookie exchange.  The default SHOULD be
 that the exchange is performed, however.  In addition, the server MAY
 choose not to do a cookie exchange when a session is resumed or, more
 generically, when the DTLS handshake uses a PSK-based key exchange.
+Servers which process 0-RTT requests and send 0.5-RTT responses
+without a cookie exchange should
```
martinthomson, line 961: SHOULD?

Isn't the alternative here to acknowledge the possibility that 0.5-RTT response are sent to an address that has not demonstrated a willingness to communicate.

draft-ietf-tls-dtls13.md
```diff
@@ -956,6 +957,10 @@ configured not to perform a cookie exchange.  The default SHOULD be
...
+store the client's IP in the cookie and limit the size of their
```
martinthomson, line 962: IMPORTANT: you just said that no cookie exchange is involved.  I think that you mean session ticket.

```suggestion
store the client's IP in the session ticket and limit the size of their
```

martinthomson (changes requested):
There is an error in the second addition here, I think.

draft-ietf-tls-dtls13.md
ekr: Do you have proposed text.

draft-ietf-tls-dtls13.md
ekr: yeah, clearly it's getting too late in the day for me to edit.

draft-ietf-tls-dtls13.md
```diff
@@ -956,6 +957,10 @@ configured not to perform a cookie exchange.  The default SHOULD be
 that the exchange is performed, however.  In addition, the server MAY
 choose not to do a cookie exchange when a session is resumed or, more
 generically, when the DTLS handshake uses a PSK-based key exchange.
+Servers which process 0-RTT requests and send 0.5-RTT responses
+without a cookie exchange should
+store the client's IP in the session ticket and limit the size of their
+responses in order to avoid amplification attacks.
```
martinthomson, line 963:
```suggestion
Servers which process 0-RTT requests and send 0.5-RTT responses
without a cookie exchange risk being used in an amplification attack
if the size of outgoing greatly exceeds the size of those that are received.
A server SHOULD limit the amount of data it sends toward a client address
before it verifies that the client is able to receive data at that address.
A client address is valid after a cookie exchange or handshake completion.
A server MAY apply a higher limit based on heuristics, such as if the client
uses the same IP address as a recent connection.
```

chris-wood: This suggested text reads fine to me. Do we want to also include @ekr's suggestion that the ticket store the client's address, if only as a way of checking "if the client uses the same IP address as a recent connection"?

chris-wood (approved):
This LGTM pending @martinthomson's suggestion and possibly a small tweak on top. This one will require consensus, too.

draft-ietf-tls-dtls13.md
```diff
@@ -956,6 +957,14 @@ configured not to perform a cookie exchange.  The default SHOULD be
...
 choose not to do a cookie exchange when a session is resumed or, more
 generically, when the DTLS handshake uses a PSK-based key exchange.
+Servers which process 0-RTT requests and send 0.5-RTT responses
+without a cookie exchange risk being used in an amplification attack
+if the size of outgoing greatly exceeds the size of those that are received.
+A server SHOULD limit the amount of data it sends toward a client address
+before it verifies that the client is able to receive data at that address.
+A client address is valid after a cookie exchange or handshake completion.
+A server MAY apply a higher limit based on heuristics, such as if the client
+uses the same IP address as a recent connection.
```
hannestschofenig, line 967: I wonder whether it makes sense to put out that this is not the regular amplification attack because the 0-RTT message is authenticated. While there is a risk of replay attack here, this assumes that 
(a) the attacker has been able to observe the initial 0-RTT request, or 
(b) the attacker is the endpoint transmitting the 0-RTT request itself. 

In case (b), if the endpoint was authenticated in the initial handshake then it is fairly easy to determine the source of the attack. In a classical amplification attack the adversary would probably also modify the source IP address to conceal its location, which does not help much here.  

Am I seeing this wrong?

ekr: No I think you're right, but I think the text covers it reasonably well.

draft-ietf-tls-dtls13.md
```diff
@@ -956,6 +957,14 @@ configured not to perform a cookie exchange.  The default SHOULD be
 that the exchange is performed, however.  In addition, the server MAY
 choose not to do a cookie exchange when a session is resumed or, more
 generically, when the DTLS handshake uses a PSK-based key exchange.
+Servers which process 0-RTT requests and send 0.5-RTT responses
+without a cookie exchange risk being used in an amplification attack
+if the size of outgoing greatly exceeds the size of those that are received.
```
kaduk, line 962:
```suggestion
if the size of outgoing messages greatly exceeds the size of those that are received.
```

kaduk (commented):
Generally fine; just a nit

draft-ietf-tls-dtls13.md
```diff
@@ -882,9 +882,9 @@ cookie technique used by Photuris {{RFC2522}} and IKE {{RFC7296}}.  When
 the client sends its ClientHello message to the server, the server
 MAY respond with a HelloRetryRequest message. The HelloRetryRequest message,
 as well as the cookie extension, is defined in TLS 1.3.
-The HelloRetryRequest message contains a stateless cookie
-(see {{TLS13}}; section 4.2.2).
-The client MUST retransmit the ClientHello
+The HelloRetryRequest message contains a stateless cookie (see
+{{RFC8446}}; Section 4.2.2).
```
chris-wood, line 886:
```suggestion
{{!TLS13}}; Section 4.2.2).
```

PR #171, opened 2020-12-07
Clarify buffering and retransmission requirements

ekr:
Don't require always buffering and require that retransmission be identical.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1129,7 +1129,7 @@ a degenerate case with fragment_offset=0 and fragment_length=length.
 Each range MUST be delivered in a single UDP datagram.
 
 When a DTLS implementation receives a handshake message fragment corresponding
-to the next expected handshake message sequence, it
+to the next expected handshake message sequence number, it
 MUST buffer it until it has the entire handshake message. DTLS
 implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1128,11 +1128,15 @@ the length field of the original message.  An unfragmented message is
 a degenerate case with fragment_offset=0 and fragment_length=length.
 Each range MUST be delivered in a single UDP datagram.
 
-When a DTLS implementation receives a handshake message fragment, it
-MUST buffer it until it has the entire handshake message.  DTLS
+When a DTLS implementation receives a handshake message fragment corresponding
+to the next expected handshake message sequence, it
+MUST buffer it until it has the entire handshake message. DTLS
 implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
-fragment sizes if the PMTU estimate changes.
+fragment sizes if the PMTU estimate changes. Handshake message bytes
+MUST NOT change upon retransmissions. Implementations MAY check
+that retransmitted bytes are identical and SHOULD abort the handshake
+with an "illegal_parameter" alert.
 
 Note that as with TLS, multiple handshake messages may be placed in
 the same DTLS record, provided that there is room and that they are
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1133,8 +1133,8 @@ to the next expected handshake message sequence number, it
 MUST buffer it until it has the entire handshake message. DTLS
 implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1128,11 +1128,15 @@ the length field of the original message.  An unfragmented message is
...
+When a DTLS implementation receives a handshake message fragment corresponding
+to the next expected handshake message sequence, it
+MUST buffer it until it has the entire handshake message. DTLS
 implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
-fragment sizes if the PMTU estimate changes.
+fragment sizes if the PMTU estimate changes. Handshake message bytes
+MUST NOT change upon retransmissions. Implementations MAY check
+that retransmitted bytes are identical and SHOULD abort the handshake
+with an "illegal_parameter" alert.
```
martinthomson, line 1139:
```suggestion
with an "illegal_parameter" alert if the value of a byte changes.
```

martinthomson (approved):

draft-ietf-tls-dtls13.md
```diff
@@ -1128,11 +1128,15 @@ the length field of the original message.  An unfragmented message is
 a degenerate case with fragment_offset=0 and fragment_length=length.
 Each range MUST be delivered in a single UDP datagram.
 
-When a DTLS implementation receives a handshake message fragment, it
-MUST buffer it until it has the entire handshake message.  DTLS
+When a DTLS implementation receives a handshake message fragment corresponding
+to the next expected handshake message sequence, it
```
chris-wood, line 1132:
```suggestion
to the next expected handshake message sequence number, it
```

draft-ietf-tls-dtls13.md
```diff
@@ -1128,11 +1128,15 @@ the length field of the original message.  An unfragmented message is
...
+MUST buffer it until it has the entire handshake message. DTLS
 implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
-fragment sizes if the PMTU estimate changes.
+fragment sizes if the PMTU estimate changes. Handshake message bytes
+MUST NOT change upon retransmissions. Implementations MAY check
```
chris-wood, line 1137:
```suggestion
fragment sizes if the PMTU estimate changes. Senders MUST NOT change 
handshake message bytes upon retransmission. Receivers MAY check
```

chris-wood (approved):

draft-ietf-tls-dtls13.md
```diff
@@ -1128,11 +1128,15 @@ the length field of the original message.  An unfragmented message is
...
-When a DTLS implementation receives a handshake message fragment, it
-MUST buffer it until it has the entire handshake message.  DTLS
+When a DTLS implementation receives a handshake message fragment corresponding
+to the next expected handshake message sequence number, it
+MUST buffer it until it has the entire handshake message. DTLS
 implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
-fragment sizes if the PMTU estimate changes.
+fragment sizes if the PMTU estimate changes. Senders MUST NOT change 
+handshake message bytes upon retransmission. Receivers MAY check
```
hannestschofenig, line 1137: What handshake message bytes are you referring to here?

draft-ietf-tls-dtls13.md
```diff
@@ -1128,11 +1128,15 @@ the length field of the original message.  An unfragmented message is
 a degenerate case with fragment_offset=0 and fragment_length=length.
 Each range MUST be delivered in a single UDP datagram.
 
-When a DTLS implementation receives a handshake message fragment, it
-MUST buffer it until it has the entire handshake message.  DTLS
+When a DTLS implementation receives a handshake message fragment corresponding
+to the next expected handshake message sequence number, it
+MUST buffer it until it has the entire handshake message. DTLS
```
hannestschofenig, line 1133: Are you making this extra clarification to keep the buffer size smaller? 

I think you are saying that when a client receives a fragment of a ServerHello message and a fragment of an EncryptedExtensions message then it needs to buffer the ServerHello message fragment (assuming that's the next message it has not yet received) while it may discard the fragment of the EncryptedExtensions message. Correct?

draft-ietf-tls-dtls13.md
kaduk: This is just trying to reiterate that retransmission is literally just retransmitting the content, with no changes.  I don't think there are any specific bytes that one might be especially tempted to change.

draft-ietf-tls-dtls13.md
ekr: That's correct. This has been the rule for some time, it's just that we managed to provide contradictory guidance.

PR #172, opened 2020-12-07
Relax age out text

ekr:
You don't want to age out before you get epoch 3 because otherwise you might be dropping when you have handshake packets.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1148,7 +1148,7 @@ from the wire and the handshake transcript: because DTLS
 records have epochs, EndOfEarlyData is not necessary to determine
 when the early data is complete, and because DTLS is lossy,
 attackers can trivially mount the deletion attacks that EndOfEarlyData
-prevents in TLS. Servers SHOULD NOT accept epoch 1 data after the first epoch 3 record
+prevents in TLS. Servers SHOULD NOT accept records from epoch 1 indefinitely once they are able to process records from epoch 3. Though reordering of IP packets can result in records from epoch 1 arriving after records from epoch 3, this is not likely to persist for very long relative to the round trip time. Servers could discard epoch 1  keys after the first epoch 3 data arrives, or retain keys for processing epoch 1 data for a short period.
 is received. (See {{dtls-epoch}} for the definitions of each epoch.)
 
 
@@ -2397,4 +2397,3 @@ In addition, we would like to thank:
 # Acknowledgements
 
 We would like to thank Jonathan Hammell for his review comments.
-
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1148,9 +1148,7 @@ from the wire and the handshake transcript: because DTLS
 records have epochs, EndOfEarlyData is not necessary to determine
 when the early data is complete, and because DTLS is lossy,
 attackers can trivially mount the deletion attacks that EndOfEarlyData
-prevents in TLS. Servers SHOULD aggressively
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1148,9 +1148,7 @@ from the wire and the handshake transcript: because DTLS
 records have epochs, EndOfEarlyData is not necessary to determine
 when the early data is complete, and because DTLS is lossy,
 attackers can trivially mount the deletion attacks that EndOfEarlyData
-prevents in TLS. Servers SHOULD aggressively
-age out the epoch 1 keys upon receiving the first epoch 2 record
-and SHOULD NOT accept epoch 1 data after the first epoch 3 record
+prevents in TLS. Servers SHOULD NOT accept epoch 1 data after the first epoch 3 record
 is received. (See {{dtls-epoch}} for the definitions of each epoch.)
```
martinthomson, line 1152: So the original advise was too strong, but I think that this is still too strong.

Early data has been accepted, the risk exists.  Being tolerant of packet ordering is a good practice.  I don't see why this couldn't just be:

> Servers SHOULD NOT accept records from epoch 1 indefinitely once they are able to process records from epoch 3.  Though reordering of IP packets can result in records from epoch 1 arriving after records from epoch 3, this is not likely to persist for very long relative to the round trip time.  Servers could discard epoch 1 data that arrives after the first epoch 3 data, or retain the ability to process epoch 1 data for a short period.

ekr: I don't see the purpose here as trying to limit damage. The complete handshake retrospectively blesses the 0-RTT data I believe. I see the purpose as just "age out keys fast". But I think your text is fine too. @chris-wood ?

martinthomson: Yes, the goal is to encourage people not to keep 0-RTT keys too long, because they might not have PCS.  The replay risk is essentially a non-issue, except to the extent that it might already have been replayed.

chris-wood: @martinthomson's suggested text seems in line with how QUIC deals with this reordering issue, so I'd vote to take his suggestion.

draft-ietf-tls-dtls13.md
```diff
@@ -1148,9 +1148,7 @@ from the wire and the handshake transcript: because DTLS
 records have epochs, EndOfEarlyData is not necessary to determine
 when the early data is complete, and because DTLS is lossy,
 attackers can trivially mount the deletion attacks that EndOfEarlyData
-prevents in TLS. Servers SHOULD aggressively
-age out the epoch 1 keys upon receiving the first epoch 2 record
-and SHOULD NOT accept epoch 1 data after the first epoch 3 record
+prevents in TLS. Servers SHOULD NOT accept epoch 1 data after the first epoch 3 record
```
chris-wood, line 1151:
```suggestion
prevents in TLS. Servers SHOULD NOT accept records from epoch 1 indefinitely once they are able to process records from epoch 3. Though reordering of IP packets can result in records from epoch 1 arriving after records from epoch 3, this is not likely to persist for very long relative to the round trip time. Servers could discard epoch 1  keys after the first epoch 3 data arrives, or retain keys for processing epoch 1 data for a short period.
```

chris-wood (changes requested):

hannestschofenig:
For my understanding: what is the harm of keeping the 0-RTT keys around till the end of the handshake?

ekr:
"Yes, the goal is to encourage people not to keep 0-RTT keys too long, because they might not have PCS. The replay risk is essentially a non-issue, except to the extent that it might already have been replayed."

PR #177, opened 2020-12-07
Records with bogus epochs should be discarded, not generate alerts

ekr:


draft-ietf-tls-dtls13.md
```diff
@@ -1611,12 +1611,10 @@ protocol exchange to allow identification of the correct cipher state:
...
-message. Implementations that receive a payload with an epoch value
-for which no corresponding cipher state can be determined MUST
-generate a "unexpected_message" alert. For example, if a client incorrectly
-uses epoch value 5 when sending early application data in a 0-RTT
-exchange. A server will not be able to compute the appropriate keys
-and will therefore have to respond with an alert.
+message. Implementations that receive a records with an epoch value
+for which no corresponding cipher state can be determined SHOULD
+handle it in the same fashion as it handles records which
+cannot be deprotected.
```
chris-wood, line 1617:
```suggestion
message. Implementations that receive a record with an epoch value
for which no corresponding cipher state can be determined SHOULD
handle it as a record which fails deprotection.
```

chris-wood (approved):
Flagging as consensus-needed.

hannestschofenig:
What was the motivation for this change?

kaduk:
> What was the motivation for this change?

The idea was to bring us in line with the normal handling for invalid records, which is to ignore them rather than send an alert and tear things down.  DTLS associations are supposed to be robust against injected packets.

I suspect that there is also some remnant of the previous state of affairs when just incrementing the epoch was all that was needed to generate new keys, as opposed to requiring KeyUpdate+ACK, though that's not directly related to making this change.

PR #179, opened 2020-12-07
Rcid excessive

ekr:
Fix a deadlock identified by @kaduk"

This seems like it sets up a deadlock scenario.  Consider peers A and B;
A sends RequestConnectionId with num_cids=200 and B considers this
request excessive, so ACKs it but sends no NewConnectionId in response.
A is prohibited from sending another RequestConnection Id to indicate
that it really needs new CIDs, since the requested 200 have not arrived
yet; thus, B could end up not sending any NewConnectionIds even though A
is essentially blocking on getting them.  Unless "unfulfilled" is
supposed to mean "not ACKed"?

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1917,8 +1917,9 @@ NewConnectionId with usage "cid_spare" containing num_cid CIDs soon as
 possible.  Endpoints MUST NOT send a RequestConnectionId message
 when an existing request is still unfulfilled; this implies that
 endpoints needs to request new CIDs well in advance.  An endpoint MAY
-ignore requests, which it considers excessive (though they MUST be
-acknowledged as usual).
+handle requests which it considers excessive by responding with
+a NewConnectionId message containing fewer than num_cid CIDs,
+including no CIDs at all.
 
 Endpoints MUST NOT send either of these messages if they did not negotiate a
 CID. If an implementation receives these messages when CIDs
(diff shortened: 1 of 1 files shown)


ekr:
@martinthomson

martinthomson (approved):

chris-wood (approved):
This one requires consensus.

draft-ietf-tls-dtls13.md
```diff
@@ -1917,8 +1917,10 @@ NewConnectionId with usage "cid_spare" containing num_cid CIDs soon as
 possible.  Endpoints MUST NOT send a RequestConnectionId message
 when an existing request is still unfulfilled; this implies that
 endpoints needs to request new CIDs well in advance.  An endpoint MAY
-ignore requests, which it considers excessive (though they MUST be
-acknowledged as usual).
+handle requests which it considers excessive by responding with
```
hannestschofenig, line 1920:
```suggestion
handle requests, which it considers excessive, by responding with
```

draft-ietf-tls-dtls13.md
```diff
@@ -1917,8 +1917,10 @@ NewConnectionId with usage "cid_spare" containing num_cid CIDs soon as
...
+a NewConnectionId message containing fewer than num_cid CIDs,
+including no CIDs at all. Endpoints MAY handle an excessive number
+of RequestConnectionId messages by terminating the connection.
```
hannestschofenig, line 1923: Hmmm. This sounds like a fragile solution to me unless we introduce a new error here to give the peer some idea why the connection was closed.

ekr: Well, we could add a new alert: "excessive_rcid_requests"

ekr: @chris-wood WDYT?

chris-wood: Alerts are cheap, so I'd be fine with that.

chris-wood: @ekr can you update this PR with a new dedicated alert?

ekr:
Done.

PR #180, opened 2020-12-07
Alerts are unreliable

ekr:


martinthomson (approved):

chris-wood (approved):
This one needs consensus, but should certainly be non-controversial.

draft-ietf-tls-dtls13.md
```diff
@@ -1492,7 +1492,9 @@ which would ordinarily issue an alert SHOULD generate a new alert
 message if the offending record is received again (e.g., as a
 retransmitted handshake message).  Implementations SHOULD detect when
 a peer is persistently sending bad messages and terminate the local
-connection state after such misbehavior is detected.
+connection state after such misbehavior is detected. Note that alerts
+are not reliably transmitted; implementation SHOULD NOT depend on
+receiving alerts in order to signal errors or connection closure.
```
hannestschofenig, line 1497: I am OK repeating text but I wonder whether the first part of this section doesn't already indicate that alerts are not reliably transmitted. 

I guess implementers had problems with this aspect and hence we are adding the text. Correct?

ekr: Or at least there was confusion

PR #194, opened 2020-12-28
Forbid going from an empty CID to a non-empty CID

### LINKED ISSUE #186: When can you send NCID and RCID
It's possible to negotiate the CID extension but that you want to receive an empty CID (this is how you get asymmetrical CID lengths). The current text says:

> Endpoints MUST NOT send either of these messages if they did not negotiate a
CID. If an implementation receives these messages when CIDs
were not negotiated, it MUST abort the connection with an unexpected_message
alert.

But does this mean that (for instance) you can negotiate receiving an empty CID and then switch to a non-empty? I am tempted to say "no" as I recall QUIC did.

ekr:
Closes #186.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1976,7 +1976,7 @@ Client                                             Server
 If the client and server have negotiated the "connection_id"
 extension {{I-D.ietf-tls-dtls-connection-id}}, either side
 can send a new CID which it wishes the other side to use
-in a NewConnectionId message.
+in a NewConnectionId message. 
 
 ~~~
 %%% Connection ID Management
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -2005,8 +2005,13 @@ MUST be used immediately for all future records. If it is set to
 Endpoints SHOULD use receiver-provided CIDs in the order they were provided.
 Endpoints MUST NOT have more than one NewConnectionId message outstanding.
 
-If the client and server have negotiated the "connection_id" extension,
-either side can request a new CID using the RequestConnectionId message.
+Implementations which either not negotiated the "connection_id" extension
```
kaduk, line 2008:
```suggestion
Implementations which either did not negotiate the "connection_id" extension
```

draft-ietf-tls-dtls13.md
```diff
@@ -1976,7 +1976,7 @@ Client                                             Server
 If the client and server have negotiated the "connection_id"
 extension {{I-D.ietf-tls-dtls-connection-id}}, either side
 can send a new CID which it wishes the other side to use
-in a NewConnectionId message.
+in a NewConnectionId message.
```
chris-wood, line 1979:
```suggestion
in a NewConnectionId message.
```

chris-wood (approved):

PR #195, opened 2020-12-28
Clarify mixing sequence and length settings on the same connection

### LINKED ISSUE #162: Clarify dynamically changing length and sequence number
Andy Cunningham wrote

> The spec is very specific on the content type fields for C(CID) and EE(Epoch bits). What about S and L? Can the sequence number encoding dynamically change while using the DTLS connection? Same question for length field? Can the connection send packets with and without the length fields (i.e. single datagrams vs multiple datagrams)?  

Text would not hurt in the draft to explain this.

ekr:
Closes #162.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -375,6 +375,8 @@ C:
 S:
 : The S bit (0x08) indicates the size of the sequence number.
   0 means an 8-bit sequence number, 1 means 16-bit.
+   Implementations MAY mix sequence numbers of different lengths
+   on the same connection.
 
 L:
 : The L bit (0x04) is set if the length is present.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -437,7 +439,8 @@ record consumes the entire rest of the datagram in the lower
 level transport. In this case it is not possible to have multiple
 DTLSCiphertext format records without length fields in the same datagram.
 Omitting the length field MUST only be used for the last record in a
-datagram.
+datagram. Implementations MAY mix records with and without length
+fields on the same connection.
```
kaduk, line 443: (I has assumed that "omitting the length only works for the last record in a datagram" implies that you can have a record with length and a record without length in the same datagram, and thus on the same connection, but don't mind saying it explicitly.)

kaduk (approved):
Chairs, please opine on whether this should get confirmed on the list.

chris-wood (approved):

PR #196, opened 2020-12-29
Use DTLS style version values, even for DTLS 1.3

### LINKED ISSUE #170: DTLS version number
@kaduk writes
We specify a ClientHello.legacy_version = {254,253}, but we seem to be
inheriting the unmodified TLS 1.3 ServerHello, complete with
ServerHello.legacy_version = 0x0303.  That seems problematic, since the
legacy DTLS 1.2 ServerHello would use the expected {254,253} like the
ClientHello.

and later

                                                                  In
      DTLS 1.3, the client indicates its version preferences in the
      "supported_versions" extension (see Section 4.2.1 of [TLS13]) and
      the legacy_version field MUST be set to {254, 253}, which was the
      version number for DTLS 1.2.  The version fields for DTLS 1.0 and
      DTLS 1.2 are 0xfeff and 0xfefd (to match the wire versions) but
      the version field for DTLS 1.3 is 0x0304.

It seems like reusing 0x0304 will make implementations more complex for
little gain -- it's common to want to, e.g., compare (D)TLS versions to
see which are greater.  OpenSSL does that with macros like:

/*
 * DTLS version numbers are strange because they're inverted. Except for
 * DTLS1_BAD_VER, which should be considered "lower" than the rest.
 */
# define dtls_ver_ordinal(v1) (((v1) == DTLS1_BAD_VER) ? 0xff00 : (v1))
# define DTLS_VERSION_GT(v1, v2) (dtls_ver_ordinal(v1) < dtls_ver_ordinal(v2))
# define DTLS_VERSION_GE(v1, v2) (dtls_ver_ordinal(v1) <= dtls_ver_ordinal(v2))
# define DTLS_VERSION_LT(v1, v2) (dtls_ver_ordinal(v1) > dtls_ver_ordinal(v2))
# define DTLS_VERSION_LE(v1, v2) (dtls_ver_ordinal(v1) >= dtls_ver_ordinal(v2))

which would have to grow another case of indirection to handle this,
whereas making TLS1_3_VERSION and DTLS1_3_VERSION have the same
numerical value doesn't seem to help the code at all.

ekr:
Closes #170.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1102,8 +1102,8 @@ legacy_version:
   extension (see Section 4.2.1 of {{!TLS13}}) and the
   legacy_version field MUST be set to {254, 253}, which was the version
   number for DTLS 1.2. The supported_versions entries for DTLS 1.0 and DTLS 1.2 are
-  0xfeff and 0xfefd (to match the wire versions) but the version field
-  for DTLS 1.3 is 0x0304.
+  0xfeff and 0xfefd (to match the wire versions). The value 0xfefc is used
+  to indicate DTLS 1.3.
 
 random:
 : Same as for TLS 1.3.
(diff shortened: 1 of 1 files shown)


ekr:
@kaduk @davidben

kaduk (approved):
This works for me, thanks.

chris-wood (approved):

davidben (approved):

Issue #159, opened 2020-11-11
Add text on how DTLS 1.2 and 1.3 is differentiated on the wire


hannestschofenig:
It would also be good to add a note on how we tell the difference between DTLS 1.2 and 1.3.
Version information is carried in various places and due to the use of the CID the multiplexing got more complex between 1.2 and 1.3.

PR #197, opened 2020-12-29
Describe how to distinguish DTLS 1.2 and DTLS 1.3 connections. Fixes …

ekr:
…#159

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -368,8 +368,8 @@ Fixed Bits:
 : The three high bits of the first byte of the DTLSCiphertext header are set to
   001. This ensures that the value will fit within the DTLS region when
   multiplexing is performed as described in {{?RFC7983}}. It also ensures
-  that it is possible to distinguish encrypted DTLS 1.3 records from
-  encrypted DTLS 1.2 records when they are carried on the same host/port quartet;
+  that distinguishing encrypted DTLS 1.3 records from encrypted DTLS 1.2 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -367,7 +367,11 @@ The DTLSCiphertext header is tightly bit-packed, as shown below:
 Fixed Bits:
 : The three high bits of the first byte of the DTLSCiphertext header are set to
   001. This ensures that the value will fit within the DTLS region when
-  multiplexing is performed as described in {{?RFC7983}}.
+  multiplexing is performed as described in {{?RFC7983}}. It also ensures
+  that it is possible to distinguish encrypted DTLS 1.3 records from
+  encrypted DTLS 1.2 records when they are carried on the same host/port quartet;
```
kaduk, line 372: (Do we care about "host/port quartet" vs. "5-tuple"?  I don't think I do...)

kaduk (approved):
This WFM

ekr:
I don't.

draft-ietf-tls-dtls13.md
chris-wood, line 372:
```suggestion
  multiplexing is performed as described in {{?RFC7983}}. It also ensures
  that distinguishing encrypted DTLS 1.3 records from encrypted DTLS 1.2 
  records is possible when they are carried on the same host/port quartet;
```

chris-wood (approved):

PR #201, opened 2021-01-07
Removed redundant statement

hannestschofenig:
Removed "Records MUST NOT span datagrams." because, in my opinion, the following sentence "Each DTLS record MUST fit within a single datagram. " in the paragraph before makes the same statement.


ekr:
I don't think these are entirely redundant: you could make a record that fit and then split it anyway. I suggest we leave it.

chris-wood:
Agreed. @hannestschofenig, I think we should close this.

chris-wood (approved):

PR #202, opened 2021-01-07
Added text regarding the legacy_session_id field

hannestschofenig:
It is not the same as TLS 1.3, as previously stated, because DTLS 1.3 does not support the compatibility mode.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1113,7 +1113,11 @@ random:
 : Same as for TLS 1.3.
 
 legacy_session_id:
-: Same as for TLS 1.3.
+: Versions of TLS and DTLS before version 1.3 supported a "session resumption" 
+feature which has been merged with pre-shared keys in version 1.3.  A client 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1113,7 +1113,11 @@ random:
 : Same as for TLS 1.3.
 
 legacy_session_id:
-: Same as for TLS 1.3.
+: Versions of TLS and DTLS before version 1.3 supported a "session resumption" 
+feature which has been merged with pre-shared keys in version 1.3.  A client 
+which has a cached session ID set by a pre-DTLS 1.3 server SHOULD set this
+field to that value. Otherwise, it MUST be set as a zero-length vector
+(i.e., a zero-valued single byte length field).
```
ekr, line 1120: This is a good change but probably needs consensus. @chris-wood: ^^?

chris-wood (approved):

kaduk (approved):
It seems a natural consequence of not having compatibility mode, but is probably worth running by the list just to cover all bases.

PR #215, opened 2021-02-08
fix duplicate TLS13 ref

kaduk:
I added this with naive editing that forgot we use [TLS13] as the tag for RFC 8446.


ekr:
This is failing CI because now the 8446 reference is missing. I'll fix it later unless you do.

kaduk:
It appears that I am confused about how the toolchain works and you should probably fix it.  Sorry for the trouble.

PR #226, opened 2021-04-06
Updates to address Martin Duke's DISCUSS

ekr:
- Change the default retransmission timer to 1s and
  allow people to do otherwise if they have side knowledge.

- Cap any given flight to 10 records

- Don't re-set the timer to the initial value but to 1.5
  times the measured RTT.

- A bunch more clarity about the reliability algorithms
  and timers (including changing reset to re-arm)

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1475,14 +1475,14 @@ and enters the WAITING state.
 There are four ways to exit the WAITING state:
 
 1. The retransmit timer expires: the implementation transitions to
-   the SENDING state, where it retransmits the flight, resets the
-   retransmit timer, and returns to the WAITING state.
+   the SENDING state, where it retransmits the flight, adjusts and re-arms the
+   retransmit timer (see {{timer-values}}), and returns to the WAITING state.
 
 2. The implementation reads an ACK from the peer: upon receiving
    an ACK for a partial flight (as mentioned in {{sending-acks}}),
    the implementation transitions
    to the SENDING state, where it retransmits the unacked portion
-   of the flight, resets the retransmit timer, and returns to the
+   of the flight, adjusts and re-arms the retransmit timer, and returns to the
    WAITING state. Upon receiving an ACK for a complete flight,
    the implementation cancels all retransmissions and either
    remains in WAITING, or, if the ACK was for the final flight,
@@ -1490,7 +1490,7 @@ There are four ways to exit the WAITING state:
 
 3. The implementation reads a retransmitted flight from the peer: the
    implementation transitions to the SENDING state, where it
-   retransmits the flight, resets the retransmit timer, and returns
+   retransmits the flight, adjusts and re-arms the retransmit timer, and returns
    to the WAITING state.  The rationale here is that the receipt of a
    duplicate message is the likely result of timer expiry on the peer
    and therefore suggests that part of one's previous flight was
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -1556,7 +1568,8 @@ is to send out only part of the flight, sending more when
 messages are acknowledged. DTLS offers a number of mechanisms
 for minimizing the size of the certificate message, including
 the cached information extension {{?RFC7924}} and certificate
-compression {{?RFC8879}}.
+compression {{?RFC8879}}. DTLS stacks MUST NOT send flights
```
hanno-arm, line 1571: The mailing list discussion recommended a SHOULD NOT, as far as I understand. It's not clear to me how a MUST NOT could be enforced in practice, if the e.g. the certificate chain is very long.

The issue here might be an overloading of the term "flight": As I read the paragraph, you use it not in the sense defined (https://www.ietf.org/archive/id/draft-ietf-tls-dtls13-41.html#name-dtls-handshake-flights) but in the sense of the part of the current flight which is currenly attempted to be transmitted. Using this latter definition, a MUST NOT as above could indeed be enforced, but ten then I think a different term should be used.

draft-ietf-tls-dtls13.md
```diff
@@ -1524,27 +1524,39 @@ flight immediately, shortcutting the retransmission timer.
...
+be retransmitted, at which time the value SHOULD be reset
+to approximately 1.5 times the measured round trip time for that
+message. After a long period of idleness, no less
 than 10 times the current timer value, implementations MAY reset the
 timer to the initial value.
 
+Note that because retransmission for handshake and not dataflow, the effect on
+congestion of shorter timeouts is smaller than in generic protocols
+such as TCP or QUIC. Experience with DTLS 1.2, which uses a
+simpler "retransmit everything on timeout" approach has not shown
```
hannestschofenig, line 1557:
```suggestion
simpler "retransmit everything on timeout" approach, has not shown
```

draft-ietf-tls-dtls13.md
```diff
@@ -1524,27 +1524,39 @@ flight immediately, shortcutting the retransmission timer.
...
-for time-sensitive applications.  Because DTLS only uses
-retransmission for handshake and not dataflow, the effect on
-congestion should be minimal.
+Unless they have extra information about the
+round trip time, implementations SHOULD use an initial timer value
+of 1000 ms and double the value at each retransmission, up to no less than 60 seconds
+(the RFC 6298 {{RFC6298}} maximum). Application specific profiles, MAY recommend shorter
+or longer timer values. For instance:
+
+* Restricted profiles such as IoT MAY specify longer timeouts.
```
hannestschofenig, line 1533:
```suggestion
* Profiles for specific deployment environments, such as in low-power, multi-hop mesh network as used in some Internet of Things (IoT) networks, MAY specify longer timeouts. See {{?I-D.ietf-uta-tls13-iot-profile}} for more information about one such DTLS 1.3 IoT profile.
```

draft-ietf-tls-dtls13.md
ekr: Agreed. Fixed.

draft-ietf-tls-dtls13.md
```diff
@@ -1532,24 +1532,44 @@ The configuration of timer settings varies with implementations and certain
...
+  MAY specify longer timeouts. See {{?I-D.ietf-uta-tls13-iot-profile}} for 
+  more information about one such DTLS 1.3 IoT profile.
+
+* Real-time protocols MAY specify shorter timeouts. It is RECOMMENDED
+  that for DTLS-SRTP {{?RFC5764}}, a default timeout of
+  400ms be used; because customer experience degrades with one-way latencies
+  of greater than 200ms, real-time deployments are less likely
+  to have long latencies. 
+
+In settings where external information (for instance from an ICE {{?RFC8445}}
```
martinduke, line 1554:
```suggestion
In settings where there is external information (for instance from an ICE {{?RFC8445}}
```

draft-ietf-tls-dtls13.md
```diff
@@ -1532,24 +1532,44 @@ The configuration of timer settings varies with implementations and certain
...
 message is transmitted and acknowledged without having to
-be retransmitted, at which time the value may be
-reset to the initial value.  After a long period of idleness, no less
+be retransmitted, at which time the value SHOULD be reset
+to approximately 1.5 times the measured round trip time for that
+message. After a long period of idleness, no less
 than 10 times the current timer value, implementations MAY reset the
 timer to the initial value.
 
+Note that because retransmission for handshake and not dataflow, the effect on
```
martinduke, line 1567:
```suggestion
Note that because retransmission is for the handshake and not dataflow, the effect on
```

martinduke (approved):
IMO this addresses my DISCUSS, though I'd like Lars's opinion as well.

Although ack-induced retransmissions are not formally reducing the send rate, as ekr pointed out the partial ack in itself will usually clear some data, so the added complexity in further limiting the retransmission is probably too much.

draft-ietf-tls-dtls13.md
```diff
@@ -1532,24 +1532,44 @@ The configuration of timer settings varies with implementations and certain
...
-(the RFC 6298 maximum). Application specific profiles, such as those
-used for the Internet of Things environment, may recommend longer
-timer values. Note that a 100 msec timer is recommended
-rather than the 3-second RFC 6298 default in order to improve latency
-for time-sensitive applications.  Because DTLS only uses
-retransmission for handshake and not dataflow, the effect on
-congestion should be minimal.
+a congested link.
+
+Unless implementations have extra information about the round trip time,
```
larseggert, line 1537:
```suggestion
Unless implementations have deployment-specific and/or external information about the round trip time, 
```

draft-ietf-tls-dtls13.md
```diff
@@ -1532,24 +1532,44 @@ The configuration of timer settings varies with implementations and certain
...
-rather than the 3-second RFC 6298 default in order to improve latency
-for time-sensitive applications.  Because DTLS only uses
-retransmission for handshake and not dataflow, the effect on
-congestion should be minimal.
+a congested link.
+
+Unless implementations have extra information about the round trip time, 
+implementations SHOULD use an initial timer value of 1000 ms and double 
+the value at each retransmission, up to no less than 60 seconds (the 
+RFC 6298 {{RFC6298}} maximum). Application specific profiles, MAY
```
larseggert, line 1540:
```suggestion
{{RFC6298}} maximum). Application-specific profiles, MAY 
```

draft-ietf-tls-dtls13.md
```diff
@@ -1532,24 +1532,44 @@ The configuration of timer settings varies with implementations and certain
...
+* Real-time protocols MAY specify shorter timeouts. It is RECOMMENDED
+  that for DTLS-SRTP {{?RFC5764}}, a default timeout of
+  400ms be used; because customer experience degrades with one-way latencies
+  of greater than 200ms, real-time deployments are less likely
+  to have long latencies. 
+
+In settings where there is external information (for instance from an ICE {{?RFC8445}}
+handshake, or from previous connections to the same server)
+about the RTT, implementations SHOULD use that to set the retransmit
+time.
```
larseggert, line 1557: There should probably be some text about how to "use that to set the retransmit time", i.e., simply reusing that value is not what TCP does. I think Linux does something like `RTO = max(SRTT + 4 * RTTVAR, 1000ms)`. [QUIC does something similar](https://www.ietf.org/archive/id/draft-ietf-quic-recovery-34.html#section-6.2.1).

draft-ietf-tls-dtls13.md
```diff
@@ -1532,24 +1532,44 @@ The configuration of timer settings varies with implementations and certain
...
+handshake, or from previous connections to the same server)
+about the RTT, implementations SHOULD use that to set the retransmit
+time. 
 
 Implementations SHOULD retain the current timer value until a
 message is transmitted and acknowledged without having to
-be retransmitted, at which time the value may be
-reset to the initial value.  After a long period of idleness, no less
+be retransmitted, at which time the value SHOULD be reset
+to approximately 1.5 times the measured round trip time for that
```
larseggert, line 1562: Why "approximately"?

larseggert (changes requested):

draft-ietf-tls-dtls13.md
ekr: Sure. I'll make this 1.5RTT as suggested by Allman and used elsewhere in this doc,

draft-ietf-tls-dtls13.md
ekr: All of these values are approximate because timers are approximate.

PR #239, opened 2021-04-08
Clients SHOULD offer a connection ID

ekr:


ekr:
@martinduke

martinduke (approved):
This definitely helps, thanks. It doesn't completely eliminate the possibility of QUIC-LB catastrophic failure, but it makes things way better.

chris-wood (approved):
LGTM, but we should confirm on the list.

hannestschofenig:
The text is good for me.

PR #245, opened 2021-05-01
Nits

kaduk:
The middle two commits should be looked at individually.
Optional-ACK can be used (as in Figure 12) to short-circuit the retransmit timer in addition to selective acknowledgment.


kaduk:
(Looks like I still don't understand the kramdown reference behavior.)

Issue #249, opened 2021-07-14
DTLS 1.3 limits the number of packets that can be encrypted with AES-GCM to 2^40.5


emanjon:
If I have understood and calculated correctly DTLS 1.3 limits the number of
packets in a AES-GCM connection to 2^40.5 compared to 2^64 in DTLS 1.2. This
is quite a severe limitation that is not mentioned explicitly. The AEAD limits
in themselves are not a problem, but the combination of AEAD limits (2^24.5)
and a small epoch (2^16) is a problem.

I think this is major problem in some use cases of DTLS. E.g. most 3GPP 5G use
cases where DTLS or DTLS/SCTP is used for semi-permanent connections (lasting years)
and where the terminating the connection cause severe disturbances.

I did not find any discussion about this problem (but I could easily have missed it)

It is known that the limit of 2^48 packets in SRTP can be reached and 2^40.5 is
significantly smaller than this...

If the document was not in such a late stage, I would have suggested to increase the size
of the epoch to 32 bits.....

I think this and the other problems DTLS 1.3 has with semi-permanent connections has to be
solved sooner than later.

ekr:
@emanjon thanks for filing this. I hear you about the late stage, but IIRC we never represent the entire epoch on the wire in any case, so it seems like just changing it to 32 bits would be pretty easy, so why not do it now?

kaduk:
How do the chacha20 numbers compare to the AES-GCM ones?

emanjon:
@kaduk I don't think (D)TLS 1.3 puts any limits on ChaCha20 so my understanding
is understanding is that the total number of packets can be protected in the current drafts is

ChaCha20 2^64 (2^48 * 2^16)
AES-GCM 2^40.5 (2^24.5 * 2^16)
AES-GCM 2^39 (2^23 * 2^16)

emanjon:
@ekr I am all for changing this now. I made a PR #250 that change epoch from uint16 to uint32. As you say, it seems like a pretty easy change to do.

Is uint32 the right size?

ekr:
Unless you think we should do uint64.

@kaduk @chris-wood what would we have to do to make this change?

emanjon:
https://github.com/gloinul/draft-westerlund-tsvwg-dtls-over-sctp-bis/issues/63

- Magnus Westerlund thinks uint64 would be a safer bet.

[...]

PR #250, opened 2021-07-14
More epochs per connection (uint32 or uint64)

emanjon:
This PR addresses #249.

Larger epoch (e.g. 32-bit) seems quite straightforward.

- It changes the DTLSPlaintext structure which is not sent on the wire.

- It changes the RecordNumber structure used in the ACK message. This should likely contain the full epoch

- It changes the 64-bit "record_sequence_number". This PR suggest the 64 low order bits of the RecordNumber.
I think it could also be set to 0x0000 || sequence_number. I don't think epoch is needed here as the keys are
different.

Note that a 32-bit epoch still limits the number of packets in a AES-GCM connection compared to what is allowed in DTLS 1.2.
2^56.6 compared to 2^64. 2^56.5 is likely enough for all use cases but epoch could also be made even larger 2^48 or 2^64.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -324,7 +324,7 @@ meaning of the fields is unchanged from previous TLS / DTLS versions.
     struct {
         ContentType type;
         ProtocolVersion legacy_record_version;
-        uint16 epoch = 0
+        uint32 epoch = 0
         uint48 sequence_number;
         uint16 length;
         opaque fragment[DTLSPlaintext.length];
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls13.md
```diff
@@ -324,7 +324,7 @@ meaning of the fields is unchanged from previous TLS / DTLS versions.
     struct {
         ContentType type;
         ProtocolVersion legacy_record_version;
-        uint16 epoch = 0
```
kaduk, line 327: Can we actually change the DTLSPlaintext format like this (recalling that it is used for the initial ClientHello)?

I would mostly have expected to leave DTLSPlaintext unchanged since you are only going to be using it with epoch values less than four.

chris-wood: > Can we actually change the DTLSPlaintext format like this (recalling that it is used for the initial ClientHello)?

@kaduk can you elaborate on the problem here? I'm not seeing the issue.

ekr: Can't we just keep this as uint16 and just call it epoch % 2^16?

chris-wood: That seems like the simplest solution here, yeah. We can just track epoch as a uint64 internally, but only take the LSB in the plaintext here.

chris-wood: (@ekr reminded me offline that this is a shared structure with DTLS 1.2, so not changing it is best.)

kaduk: > (@ekr reminded me offline that this is a shared structure with DTLS 1.2, so not changing it is best.)

Right, that is the problem I was thinking of.

I think `epoch % 2^16` should work, especially since we don't actually use this structure with large epochs (we've moved to DTLSCiphertext by then).

chris-wood:

PR #252, opened 2021-07-16
More epochs per connection (epoch wrap)

emanjon:
Alternative solution to address issue #249 

This may be a simpler solution unless there are any security implications. DTLS 1.0 allowed epoch wrapping. DTLS 1.2 forbid epoch wrapping. I don't know the history behind this or if the reasons for forbidding epoch wrapping in DTLS 1.2 still apply to DTLS 1.3


gloinul:
This change should not be accepted until one have decided if #253 makes sense, because then non-wrapping epoch would avoid creating a risk with the key exporter.

chris-wood:

PR #255, opened 2021-08-14
Increase effective epoch width to 64 bits, keeping the serialized representation 16 bits wide

chris-wood:
Yet another alternative to #249. 

cc @emanjon @hannestschofenig

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -484,8 +484,8 @@ unpacked RecordNumber structure, as shown below:
     } RecordNumber;
 ~~~
 
-This 112-bit value is used in the ACK message. The low order 64 bits are used in
-the "record_sequence_number" input to the AEAD function.
+This 112-bit value is used in the ACK message. The low order 64 bits are used as
+the sequence number for reading and writing records; see {{TLS13, Section 5.3}}.
 
 The entire header value shown in {{hdr_examples}} (but prior to record number
 encryption, see {{rne}}) is used as as the additional data value for the AEAD
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -484,8 +484,8 @@ unpacked RecordNumber structure, as shown below:
     } RecordNumber;
 ~~~
 
-This 64-bit value is used in the ACK message as well as in the "record_sequence_number"
-input to the AEAD function.
+This 112-bit value is used in the ACK message. The low order 64 bits are used in
+the "record_sequence_number" input to the AEAD function.
 
 The entire header value shown in {{hdr_examples}} (but prior to record number
 encryption, see {{rne}}) is used as as the additional data value for the AEAD
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -313,8 +313,8 @@ also different from the DTLS 1.2 record layer.
 
 3. The DTLS epoch serialized in DTLSPlaintext is 2 octets long for compatibility
    with DTLS 1.2. However, this value is set as the least significant 2 octets
-   from the effective epoch of a connection, which is an 8 octet counter
-   incremented on every KeyUpdate. See {{seq-and-epoch}} for details.
+   of the epoch of a connection, which is an 8 octet counter incremented on every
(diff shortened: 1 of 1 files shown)


ekr:
This seems like the right change (modulo analysis) but I think the use of the term "effective" is a mistake. We already carry a truncated epoch in DTLSCiphertext, so I think all that's needed is to say that the epoch field in DTLSPlaintext is also truncated but that the actual epoch is 64 bits.

chris-wood:
> so I think all that's needed is to say that the epoch field in DTLSPlaintext is also truncated but that the actual epoch is 64 bits.

Dropped "effective," so the current PR should reflect this. 👍

draft-ietf-tls-dtls13.md
```diff
@@ -311,7 +311,12 @@ also different from the DTLS 1.2 record layer.
    the DTLSCiphertext structure have been reduced from those in previous
    versions.
 
-3. The DTLSCiphertext structure has a variable length header.
+3. The DTLS epoch serialized in DTLSPlaintext is 2 octets long for compatibility
+   with DTLS 1.2. However, this value is set as the least significant 2 octets
+   from the effective epoch of a connection, which is an 8 octet counter
```
ekr:
```suggestion
   from the connection epoch, which is an 8 octet counter
```

draft-ietf-tls-dtls13.md
```diff
@@ -350,6 +355,9 @@ legacy_record_version
   It MUST be ignored for all purposes. See {{TLS13}}; Appendix D.1
   for the rationale for this.
 
+legacy_epoch
```
ekr, line 358: I think you can drop ```legacy``` here. We don't use it for DTLSCiphertext

ekr (approved):
This looks right, again modulo analysis.

draft-ietf-tls-dtls13.md
chris-wood: Works for me!

draft-ietf-tls-dtls13.md
```diff
@@ -471,7 +479,7 @@ unpacked RecordNumber structure, as shown below:
 ~~~
 %%% Record Layer
     struct {
-        uint16 epoch;
+        uint64 epoch;
         uint48 sequence_number;
     } RecordNumber;
 ~~~
```
martinthomson, line 485: Down on line 479, it says "This 64-bit value is used in the ACK message as well as in the "record_sequence_number" input to the AEAD function.

Two problems:

1. It's no longer 64 bits.  It's 112 bits.
2. It's no longer 64 bits.  It's 112 bits.

The first is just correcting the text.

The second is that the per-record nonce construction in https://datatracker.ietf.org/doc/html/rfc8446#section-5.3 no longer works for most, if not all, modern ciphers without additional transforms (which would reduce, not improve security).

I think that the fix here is to make the epoch 48 bits rather than 64.  Still, this will need a lot more text in order to make the generation of the per-record nonce clear.  You need to point to Section 5.3 of 8446 and say what is different and why.

(I searched for "record_sequence_number" in this document and RFC 8446 and found no other mentions of the string.  I'm sure that people without context will really struggle with this.)

martinthomson (changes requested):
This doesn't work as far as I can tell.

draft-ietf-tls-dtls13.md
```diff
@@ -471,13 +479,13 @@ unpacked RecordNumber structure, as shown below:
...
 
-This 64-bit value is used in the ACK message as well as in the "record_sequence_number"
-input to the AEAD function.
+This 112-bit value is used in the ACK message. The low order 64 bits are used in
+the "record_sequence_number" input to the AEAD function.
```
martinthomson, line 488: So I don't think that this is quite enough.  Technically, it's fine, but "record_sequence_number" is not a thing.  I couldn't find a definition in this doc or RFC 8446.  Why not just cite [Section 5.3 of RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446#section-5.3)?

chris-wood: That's existing text, but we can certainly improve it here.

chris-wood: @martinthomson - done.

chris-wood:

PR #257, opened 2021-11-08
Extended epoch like tls. Fixed #249

### LINKED ISSUE #249: DTLS 1.3 limits the number of packets that can be encrypted with AES-GCM to 2^40.5

ekr:
Yet another attempt to fix #249.

This builds on Chris's PR but just omits the epoch entirely from the AEAD nonce calculation. This is more consistent with TLS and doesn't require special case reasoning about why we only need to build in the bottom 16 bits. This relies entirely on the keys being different. Needs analysis.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -484,8 +484,8 @@ unpacked RecordNumber structure, as shown below:
     } RecordNumber;
 ~~~
 
-This 112-bit value is used in the ACK message. The low order 64 bits are used in
-the "record_sequence_number" input to the AEAD function.
+This 112-bit value is used in the ACK message. The low order 64 bits are used as
+the sequence number for reading and writing records; see {{TLS13, Section 5.3}}.
 
 The entire header value shown in {{hdr_examples}} (but prior to record number
 encryption, see {{rne}}) is used as as the additional data value for the AEAD
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -484,8 +484,8 @@ unpacked RecordNumber structure, as shown below:
     } RecordNumber;
 ~~~
 
-This 64-bit value is used in the ACK message as well as in the "record_sequence_number"
-input to the AEAD function.
+This 112-bit value is used in the ACK message. The low order 64 bits are used in
+the "record_sequence_number" input to the AEAD function.
 
 The entire header value shown in {{hdr_examples}} (but prior to record number
 encryption, see {{rne}}) is used as as the additional data value for the AEAD
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -2164,6 +2164,14 @@ Client                                             Server
 ~~~
 {: #dtls-key-update title="Example DTLS Key Update"}
 
+Implementations MUST NOT send KeyUpdates that would result
+in epochs of greater than or equal to 2^{48}-1. This limits
+the probability that a 128 bit key will be repeated during
+a connection. Note, however, that even if the key repeats,
+the IV is also independently generated. In order to allow
+future relaxation of this requirement, receiving implementations
+MUST NOT enforce this rule.
+
 
 # Connection ID Updates
 
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -314,7 +314,8 @@ also different from the DTLS 1.2 record layer.
 3. The DTLS epoch serialized in DTLSPlaintext is 2 octets long for compatibility
    with DTLS 1.2. However, this value is set as the least significant 2 octets
    of the connection epoch, which is an 8 octet counter incremented on every
-   KeyUpdate. See {{seq-and-epoch}} for details.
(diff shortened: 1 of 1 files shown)


kaduk:
This seems promising.

chris-wood (approved):
Thanks -- this LGTM. We'll need to do a brief consensus call given the type of change.

draft-ietf-tls-dtls13.md
```diff
@@ -311,7 +311,15 @@ also different from the DTLS 1.2 record layer.
    the DTLSCiphertext structure have been reduced from those in previous
    versions.
 
-3. The DTLSCiphertext structure has a variable length header.
+3. The DTLS epoch serialized in DTLSPlaintext is 2 octets long for compatibility
+   with DTLS 1.2. However, this value is set as the least significant 2 octets
+   of the connection epoch, which is an 8 octet counter incremented on every
+   KeyUpdate. See {{seq-and-epoch}} for details. The sequence number is set to
+   be the low order 48 bits of the 64 bit sequence number. Plaintext records
+   MUST NOT be sent with sequence numbers that would exceedd 2^48-1, so the
```
martinthomson, line 319:
```suggestion
   MUST NOT be sent with sequence numbers that would exceed 2^48-1, so the
```

draft-ietf-tls-dtls13.md
```diff
@@ -2151,6 +2164,14 @@ Client                                             Server
 ~~~
 {: #dtls-key-update title="Example DTLS Key Update"}
 
+Implementations MUST NOT send KeyUpdates that would result
+in epochs of greater than or equal to 2^{48}-1. This limits
```
martinthomson, line 2168: This is somewhat arbitrary.  If the epoch can have 64 bit values, then why not allow for the full range of values to be used?  Is there an analysis that suggests that rekeying too often leads to this problem?

Also, I would have said >2^48-1 OR >= 2^48, but not this.  I can almost guarantee a test failure in some compliance test arising from that choice.  The test will be pointless too.

martinthomson (commented):
Many implementations of DTLS use a 16-bit integer to hold epoch.  Sometimes, that leaks into places that mean that it is hard to change.

Those of us with this problem can probably pull the same trick as here and only expose the low 16 bits, but would it be possible to implement this without allowing key updates that take epoch to 2^16?

draft-ietf-tls-dtls13.md
ekr: With a key space of 2^{128}, 2^{64} rekeys has a significant chance of key repetition.

The IV randomization is intended to address this, but I prefer to be cautious, especially as there is no realistic scenario in which > 24^{48} rekeys will be needed.

ekr: Happy to change this to >= 2^{48}

martinthomson: If we accept the premise that the nonce is effectively part of the key space, then AES-128 is still probably OK with that many rekeys.  AES-256 and ChaCha20Poly1305 have no such problem.  I'm just concerned that this is arbitrary and unexplained.

chris-wood:
@ekr this LGTM, but there's a conflict. Can you resolve prior to merging? 

@martinthomson, are you OK with the new rationale, and text that allows it to be relaxed in the future?

draft-ietf-tls-dtls13.md
```diff
@@ -1,4 +1,4 @@
----
+2^---
```
martinthomson, line 1:
```suggestion
---
```

draft-ietf-tls-dtls13.md
```diff
@@ -2151,6 +2161,17 @@ Client                                             Server
...
 
+With a 128-bit key as in AES-256, rekeying 2^64 times has a high
+probability of key reuse within a given connection. Note that even if
+the key repeats, the IV is also independently generated. In order to
+provide an extra margin of security, sending implementations MUST NOT
+allow the epoch to exceed 2^48-1. In order to allow this value to
+be changed later, receiving implementations MUST NOT
+enforce this rule. If a sending implementation receives a KeyUpdate
+with request_update set to "update_requested", it MUST NOT send
+its own KeyUpdate if that would cause it to exceed these limits.
```
martinthomson, line 2172: What can it do instead?

martinthomson (approved):
Seems OK.

PR #258, opened 2021-11-10
Allow immediate processing of handshake data, where possible

hanno-arm:
Change as agreed in https://mailarchive.ietf.org/arch/msg/tls/PLK-EW_bWIXV74nm_8gSHwu5cqk/.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1266,8 +1266,9 @@ MUST be delivered in a single UDP datagram.
 
 When a DTLS implementation receives a handshake message fragment corresponding
 to the next expected handshake message sequence number, it
-MUST buffer it until it has the entire handshake message. DTLS
-implementations MUST be able to handle overlapping fragment ranges.
+MUST buffer it until it has the entire handshake message. Where possible,
+implementations MAY instead process data immediately rather than buffering it.
+DTLS implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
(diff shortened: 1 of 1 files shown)


hanno-arm:
@ekr @chris-wood Any comments/concerns?

chris-wood (approved):
"Where possible" is a bit unclear, but I don't have any better suggestions at the moment.

kaduk (commented):
I would be interested to see an alternative proposal that moves to a "MUST NOT discard" framing.  This version of "MUST buffer ... but MAY as an alternative process" feels a bit like giving conflicting advice.

I note that the "next expected handshake message sequence number" is  important here -- fragments of subsequent messages can be discarded.

hanno-arm:
@kaduk Good idea!

What about this, @kaduk @ekr @chris-wood:
> When a DTLS implementation receives a handshake message fragment corresponding to the next expected handshake message sequence number, it MUST NOT discard it. It is RECOMMENDED that the implementation buffers such fragments until it has the entire handshake message, but it MAY also process the fragment immediately.

chris-wood:
@hanno-arm this seems better given that it avoids the "where possible" ambiguity. @ekr, what do you think?

hanno-arm:
@ekr Appreciate it's not the most important issue, but can you take a look so we can get this in one form or another?

hanno-arm:
@ekr What are your concerns with the present version?

ekr:
Well, originally I just thought my text was clearer. But now that I know that the intent is to allow out-of-order processing I have a technical objection.

chris-wood:
Closing as proposed in #261.

PR #261, opened 2021-12-01
Another take on being able to process early. Fixes #259.

### LINKED ISSUE #259: Consider rewording of handshake buffering requirement
>    When a DTLS implementation receives a handshake message fragment
   corresponding to the next expected handshake message sequence number,
   it MUST buffer it until it has the entire handshake message. 

As discussed in https://mailarchive.ietf.org/arch/msg/tls/PLK-EW_bWIXV74nm_8gSHwu5cqk/: The spec mandates buffering all handshake fragments, but some implementations (e.g. GnuTLS) choose to implement simpler reassembly schemes. It has been argued by Ekr that such should be discouraged because they aren't efficient. On the other hand, as Rich Salz remarks, the choice of reassembly mechanism isn't something the peer could tell (the effects of poor reassembly performance are more retransmissions, which could also be due to poor network reliability), so it does feel like we're overspecifying things here. Weakening the MUST to a SHOULD isn't good either as it allows dropping fragments altogether. There's a tension between the desire for a concrete operational guidance, which the current text clearly provides, and a more abstract formulation along the line of "implementations MUST use a reassembly scheme ensuring forward progress". 

As suggested by Ekr, taking this off the mailing list and continuing discussion here.

To be clear: This has no bearing on the protocol itself, it's about how much implementation details the spec should enforce.

ekr:
@kaduk here is another try at this.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1266,8 +1266,9 @@ MUST be delivered in a single UDP datagram.
 
 When a DTLS implementation receives a handshake message fragment corresponding
 to the next expected handshake message sequence number, it
-MUST buffer it until it has the entire handshake message. DTLS
-implementations MUST be able to handle overlapping fragment ranges.
+MUST process it, either by buffering it until it has the entire handshake message
+or by processing any in-order portions of the message.
+DTLS implementations MUST be able to handle overlapping fragment ranges.
 This allows senders to retransmit handshake messages with smaller
 fragment sizes if the PMTU estimate changes. Senders MUST NOT change
 handshake message bytes upon retransmission. Receivers MAY check


hanno-arm (changes requested):
As mentioned on the list, I don't like the in-order qualification. I agree it's debatable whether there will be cases where data can be processed out of order, but (see list again) it is not inconceivable, and I see no reason to exclude this here. I would therefore prefer to drop the in-order qualification.

ekr:
I did not realize that the intent of your text was to allow out of order processing. IMO that's off the fairway and
we shouldn't allow that any more than we let you reach into the TCP stack and pull out forward messages.

hanno-arm:
@ekr I don't think the comparison with TCP works - the handshake protocol is internal. We should not force a particular implementation unless there's good reason to. In this case, as explained on the list, there's potential merit of allowing out of order processing of very large handshake messages with known structure.
@chris-wood What's your opinion?

ekr:
Again, this is a last minute change during Auth48, so the bar is *much*
higher than "potential merit"

chris-wood:
This isn't motivated by security-related reasons like #257, so I agree with @ekr that this does not raise to the level of an AUTH48 change. I propose we land this PR and close out #258.

chris-wood (approved):

hanno-arm:
Thanks @chris-wood @ekr for considering this in the first place. As I understood @ekr, since this does not impact the wire-format, we can revisit once the RFC has been finalized?

chris-wood:
> we can revisit once the RFC has been finalized?

Once the RFC is finalized, it's final. Changes can be requested via the [errata process](https://www.ietf.org/about/groups/iesg/statements/processing-rfc-errata/), or via a new document.

hanno-arm:
I was referring to @ekr's comment on the list
> ... we can just do an update without changing the protocol on the wire.

ekr:
Yes. What I was saying is that you could just write an RFC that said "you can process out of order"

PR #265, opened 2022-01-12
Renaming HandshakeType names in IANA request

### LINKED ISSUE #264: Rename TLS HandshakeTypes
This I-D registers two TLS HandshakeTypes. All of the other fields in the registry use snake_case.

seanturner:
Resolves #264.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1105,6 +1105,8 @@ fragmentation, DTLS modifies the TLS 1.3 handshake header:
         end_of_early_data(5),
         hello_retry_request_RESERVED(6),
         encrypted_extensions(8),
+        request_connection_id(9),
+        new_connection_id(10),
         certificate(11),
         server_key_exchange_RESERVED(12),
         certificate_request(13),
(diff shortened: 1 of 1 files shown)


kaduk (commented):
Do we want anything in the prose about how the RequestConnectionId structure is the message contents when the HandshakeType is request_connection_id/etc.?
Also update Appendix A.2?

seanturner:
As far as referring to the snake case in the main body, I went and looked at RFC 8446 and it does not refer to the snake case.  I tend to think that if we make the change in 5.2 and A.2, that should be enough breadcrumbs for people to figure it out.

I put them in the same order as they are in the registry and moved up new_session_ticket.

Triple check this change please.

chris-wood (approved):

draft-ietf-tls-dtls13.md
```diff
@@ -1130,12 +1132,14 @@ fragmentation, DTLS modifies the TLS 1.3 handshake header:
             case client_hello:          ClientHello;
             case server_hello:          ServerHello;
             case end_of_early_data:     EndOfEarlyData;
+            case new_session_ticket:    NewSessionTicket;
             case encrypted_extensions:  EncryptedExtensions;
-            case certificate_request:   CertificateRequest;
+            case request_connection_id: RequestedConnectionId;
```
kaduk, line 1137: Shouldn't this be just "RequestConnectionId", not "Requested"?

draft-ietf-tls-dtls13.md
```diff
@@ -1130,12 +1132,14 @@ fragmentation, DTLS modifies the TLS 1.3 handshake header:
             case client_hello:          ClientHello;
             case server_hello:          ServerHello;
             case end_of_early_data:     EndOfEarlyData;
+            case new_session_ticket:    NewSessionTicket;
```
kaduk, line 1135: I see NST (4) as before EOED (5) in the registry.

kaduk: Were we planning to order these by codepoint value in the final RFC?

seanturner: Yes ... will submit a new PR now to reorder

draft-ietf-tls-dtls13.md
seanturner: Yes fixing this too.

PR #269, opened 2022-02-25
Clarify closure alert

### LINKED ISSUE #268: DTLS behavior on received closure alert
In RFC6083bis (DTLS/SCTP) we are discussing when to terminate DTLS connections. 

RFC8446 says:

  "close_notify:  This alert notifies the recipient that the sender will
   not send any more messages on this connection.  **Any data received
   after a closure alert has been received MUST be ignored."**

As far as I can see, the DTLS 1.3 draft does not say anything about closure alerts,
which according to my understanding means that the above RFC 8446 text apply to
DTLS 1.3 as well. However, this behaviour does not seem to make much sense for DTLS 1.3.
As DTLS records can be received out of order, shouldn't the DTLS behavior be:

  "Any data sent after a closure alert MUST be ignored."

ekr:
Fixes #268.

### DIFF
--- a/draft-ietf-tls-dtls13.md
+++ b/draft-ietf-tls-dtls13.md
@@ -1675,6 +1675,9 @@ connection state after such misbehavior is detected. Note that alerts
 are not reliably transmitted; implementations SHOULD NOT depend on
 receiving alerts in order to signal errors or connection closure.
 
+Any data received with an epoch/sequence number pair after
+that of a valid received closure alert MUST be ignored.
+
 ##  Establishing New Associations with Existing Parameters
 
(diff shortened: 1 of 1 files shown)


kaduk (approved):
I agree that this is the semantics we want from the alert.
Do we need to say something that specifically countermands the directive from RFC 8446 that essentially uses a time-based criterion in addition to this, though?

gloinul (approved):
Sufficient. I guess it has more impact on the implementations as you need to separate the received close and the actual final closing that discard the keys and state allowing to process the data.

hannestschofenig:
It looks like a reasonable statement to make

PR #271, opened 2022-03-05
fix msg_type's ordering and name

seanturner:
In #265 didn't quite get it right. Fixing it now so the order matches the IANA reg, and the name case matches the field name.


chris-wood (approved):

kaduk:
Thanks!

PR #294, opened 2025-03-03
Receiving ACKs Handling

hannestschofenig:
See errata 8108: https://www.rfc-editor.org/errata/eid8108


hannestschofenig:
Yes, I also think this should be tracked separately. PR #294 addresses the concrete receiver-side check from Errata 8108: an ACK must not acknowledge records from an epoch higher than the epoch in which the ACK itself was received.

The second part of David's comment is broader. It asks whether the specification should explicitly define when receivers close old epochs and stop accepting records protected under them. The current text gives general guidance to discard
records from earlier epochs or retain old keying material for reordering, but it does not define precise closure points. That seems worth a separate issue, because it affects record processing and key lifetime beyond ACK validation.

ekr (approved):
LGTM

PR #322, opened 2025-11-21
Attempt to rewrite record handling

### LINKED ISSUE #283: Imprecise definitions for handling invalid records
[Section 4.5.2](https://tlswg.org/dtls13-spec/draft-ietf-tls-dtls13.html#section-4.5.2) talks about invalid record handling in a way that is not sufficiently precise.

Take this:

> Unlike TLS, DTLS is resilient in the face of invalid records (e.g., invalid formatting, length, MAC, etc.). In general, invalid records SHOULD be silently discarded, thus preserving the association; [...]

This recommendation is not great, particularly as it attached to the preceding list, which is a little non-specific.

Quoting @dennisjackson from another context:

> I do still think the phrasing in 4.5.2 is needlessly hostile by introducing the concept of an invalid record in the first place. If we asked implementers which of the following are considered invalid records, I'd suspect they'd say "all of them":
> 
>  * A record which consists of a zero length handshake message
>  * A record which contains a fragment of an alert message (or two alert messages).
>  * A record of application data which appears between two records of a fragmented handshake message.
>  * A record which has an invalid inner content type.
>  * A record which has an invalid outer content type.
>
> I do think 4.5.2 would be significantly clearer if it didn't mention "invalid records" at all and just spoke about "errors arising from record handling".

To this point, it would be clearer if the spec were more definitive.  Records that do not authenticate would always be discarded (i.e., use MUST) and records that do authenticate, but are otherwise invalid, result in fatal alerts always (i.e., use MUST).  I don't see much value in having ambiguous language for something so important.

martinthomson:
This takes the feedback in #283 and attempts to restructure the section to be clearer about what invalid record processing means.  And what to do if it happens.

I did not take @dennisjackson's point about application data interleaved between two handshake records in the same datagram.  That seemed a bit esoteric, so I turned it into something less specific.  That might mean that sometimes people will miss that very specific corner case of bad peer.

Many more paragraphs, but a lot of the text, aside from the list and the one paragraph before and two paragraphs after it, is just copied and pasted from the existing walloftext.

Closes #283.

### DIFF
--- a/draft-ietf-tls-rfc9147bis.md
+++ b/draft-ietf-tls-rfc9147bis.md
@@ -845,24 +845,57 @@ record has been deprotected  successfully.
 
 ### Handling Invalid Records
 
-Unlike TLS, DTLS is resilient in the face of invalid records (e.g.,
-invalid formatting, length, MAC, etc.).  In general, invalid records
-SHOULD be silently discarded, thus preserving the association;
-however, an error MAY be logged for diagnostic purposes.
-Implementations which choose to generate an alert instead MUST
-generate fatal alerts to avoid attacks where the attacker
-repeatedly probes the implementation to see how it responds to
-various types of error.  Note that if DTLS is run over UDP, then any
-implementation which does this will be extremely susceptible to
-DoS attacks because UDP forgery is so easy.
-Thus, generating fatal alerts is NOT RECOMMENDED for such transports, both
-to increase the reliability of DTLS service and to avoid the risk
-of spoofing attacks sending traffic to unrelated third parties.
+Unlike TLS, DTLS can be resilient when receiving invalid records (e.g.,
+invalid formatting, length, MAC, etc.).
+
+When receiving a record that is invalid, endpoints SHOULD discard the record
+rather than generating a fatal alert. This includes only the construction of
+records and the AEAD protection on the record (if any), not the contents of
+a record that is successfully decrypted.  This includes:
+
+* Records that contain invalid header values.  For instance, records with
+  invalid values for fields other than `fragment` in `DTLSPlaintext` or
+  `DTLSCiphertext`; see {{dtls-record}}.
+
+* Records where length does not match the length of available data. For
+  instance, where the length indicates a length longer than what remains in
+  a UDP datagram.
+
+* Records with an epoch that is not currently being accepted.
+
+* Records where authenticated encryption protection cannot be removed
+  successfully.
+
+* Records that contain an invalid inner content type or unterminated
+  padding; see `DTLSInnerPlaintext` in {{dtls-record}}}.
+
+Records that are successfully decoded and decrypted can be assumed to be
+valid and errors in the contents of records MUST be regarded as originating
+from the peer.  Fatal alerts can be generated based on the content of
+records.  This includes all the unprotected records exchanged as part of the
+handshake.
+
+Note that this means that DTLS does not safeguard against denial of service
+caused by damage to handshake messages carried in unprotected records.
+
+In general, invalid records SHOULD be silently discarded, thus preserving
+the association; however, an error MAY be logged for diagnostic purposes.
+
+Generating fatal alerts is NOT RECOMMENDED for unreliable transports, both to
+increase the reliability of DTLS service and to avoid the risk of spoofing
+attacks sending traffic to unrelated third parties.
 
 If DTLS is being carried over a transport that is resistant to
 forgery (e.g., SCTP with SCTP-AUTH {{RFC6083}}), then it is safer to send alerts
 because an attacker will have difficulty forging a datagram that will
 not be rejected by the transport layer.
+
+Implementations which choose to generate an alert in response to an invalid
+record MUST generate fatal alerts to avoid attacks where the attacker
+repeatedly probes the implementation to see how it responds to various types
+of error.  Note that if DTLS is run over UDP, then any implementation which
+does this will be extremely susceptible to DoS attacks because UDP forgery
+is so easy.
 
 Note that because invalid records are rejected at a layer lower than
 the handshake state machine, they do not affect pending
--- a/draft-ietf-tls-rfc9147bis.md
+++ b/draft-ietf-tls-rfc9147bis.md
@@ -870,7 +870,7 @@ the record layer.  This includes:
 
 * Records that cannot be demultiplexed as DTLS records.
 
-Records that are successfully decoded and decrypted can be assumed to be
+Records that are successfully decoded and deprotected can be assumed to be
 valid and errors in the contents of records MUST be regarded as originating
 from the peer.  Fatal alerts can be generated based on the content of
 records.  This includes all the unprotected records exchanged as part of the
@@ -881,10 +881,6 @@ caused by damage to handshake messages carried in unprotected records.
 
 In general, invalid records SHOULD be silently discarded, thus preserving
 the association; however, an error MAY be logged for diagnostic purposes.
-
-Generating fatal alerts is NOT RECOMMENDED for unreliable transports, both to
-increase the reliability of DTLS service and to avoid the risk of spoofing
-attacks sending traffic to unrelated third parties.
 
 If DTLS is being carried over a transport that is resistant to
 forgery (e.g., SCTP with SCTP-AUTH {{RFC6083}}), then it is safer to send alerts
--- a/draft-ietf-tls-rfc9147bis.md
+++ b/draft-ietf-tls-rfc9147bis.md
@@ -845,29 +845,30 @@ record has been deprotected  successfully.
 
 ### Handling Invalid Records
 
-Unlike TLS, DTLS can be resilient when receiving invalid records (e.g.,
-invalid formatting, length, MAC, etc.).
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+  instance, where the length indicates a length longer than what remains in
+  a UDP datagram.
+
+* Records with an epoch that is not currently being accepted.
+
+* Records where authenticated encryption protection cannot be removed
+  successfully.
+
+* Records that contain an invalid inner content type or unterminated
+  padding; see `DTLSInnerPlaintext` in {{dtls-record}}}.
```
hannestschofenig:
```suggestion
  padding; see `DTLSInnerPlaintext` in {{dtls-record}}.
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
-generate fatal alerts to avoid attacks where the attacker
-repeatedly probes the implementation to see how it responds to
-various types of error.  Note that if DTLS is run over UDP, then any
-implementation which does this will be extremely susceptible to
-DoS attacks because UDP forgery is so easy.
-Thus, generating fatal alerts is NOT RECOMMENDED for such transports, both
-to increase the reliability of DTLS service and to avoid the risk
-of spoofing attacks sending traffic to unrelated third parties.
+Unlike TLS, DTLS can be resilient when receiving invalid records (e.g.,
+invalid formatting, length, MAC, etc.).
```
hannestschofenig:
```suggestion
Unlike TLS, DTLS can be resilient when receiving records that cannot be
processed at the record layer.
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+
+When receiving a record that is invalid, endpoints SHOULD discard the record
+rather than generating a fatal alert. This includes only the construction of
+records and the AEAD protection on the record (if any), not the contents of
+a record that is successfully decrypted.  This includes:
```
hannestschofenig:
```suggestion
An endpoint MUST discard a record, without updating handshake state, ACK state,
retransmission state, or replay windows, when the record cannot be processed at
the record layer.  This includes:
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+
+* Records that contain invalid header values.  For instance, records with
```
hannestschofenig:
```suggestion
* Records with invalid header values.  For instance, records with
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+  invalid values for fields other than `fragment` in `DTLSPlaintext` or
+  `DTLSCiphertext`; see {{dtls-record}}.
+
+* Records where length does not match the length of available data. For
```
hannestschofenig:
```suggestion
* Records whose length is inconsistent with the remaining datagram contents. For
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+  instance, where the length indicates a length longer than what remains in
+  a UDP datagram.
+
+* Records with an epoch that is not currently being accepted.
+
+* Records where authenticated encryption protection cannot be removed
```
hannestschofenig:
```suggestion
* Records where AEAD protection cannot be removed
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+  successfully.
+
+* Records that contain an invalid inner content type or unterminated
```
hannestschofenig:
```suggestion
* Records with an invalid inner content type or invalid; and
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+  padding; see `DTLSInnerPlaintext` in {{dtls-record}}}.
+
```
hannestschofenig, line 872: * Records that cannot be demultiplexed as DTLS records.

hannestschofenig:
Thank you Martin for addressing your own comment. I agree that discussing the differences of what constitutes an "invalid record" is an improvement. In my comments I have made some wording changes. Many of them are not important but could be easier to read.

draft-ietf-tls-rfc9147bis.md
martinthomson:
```suggestion

* Records that cannot be demultiplexed as DTLS records.

```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,58 @@ record has been deprotected  successfully.
...
+Implementations which choose to generate an alert in response to an invalid
+record MUST generate fatal alerts to avoid attacks where the attacker
+repeatedly probes the implementation to see how it responds to various types
+of error.  Note that if DTLS is run over UDP, then any implementation which
+does this will be extremely susceptible to DoS attacks because UDP forgery
+is so easy.
+
 Note that because invalid records are rejected at a layer lower than
 the handshake state machine, they do not affect pending
 retransmission timers.
```
martinthomson: I saw the first batch of suggestions and thought, "oh, typos, no problem".  But this is a complete rewrite.  Hard to comment on a suggestion of this size.  Want to take this one over as a separate PR, Hannes?  Otherwise, this is hard to comment on.

The things that I think are bad about this (and the original; it was now more than 7 months since I opened this PR) are mostly that "SHOULD NOT".  I think that's bad advice now.  What you want is robustness at the DTLS layer such that things are discarded always, unless the AEAD says that they are OK.  What SCTP AUTH might give is better reliability for the handshake.

hannestschofenig: I might have gone a bit overboard with my proposed text changes. I want your PR to get merged because it contains a lot of improvements.

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,59 @@ record has been deprotected  successfully.
...
+
+* Records where AEAD protection cannot be removed
+  successfully.
+
+* Records with an invalid inner content type or invalid; and
+  padding; see `DTLSInnerPlaintext` in {{dtls-record}}.
+
+* Records that cannot be demultiplexed as DTLS records.
+
+Records that are successfully decoded and decrypted can be assumed to be
```
ekr:
```suggestion
Records that are successfully decoded and deprotected can be assumed to be
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,59 @@ record has been deprotected  successfully.
...
 forgery (e.g., SCTP with SCTP-AUTH {{RFC6083}}), then it is safer to send alerts
 because an attacker will have difficulty forging a datagram that will
 not be rejected by the transport layer.
 
+Implementations which choose to generate an alert in response to an invalid
+record MUST generate fatal alerts to avoid attacks where the attacker
+repeatedly probes the implementation to see how it responds to various types
+of error.  Note that if DTLS is run over UDP, then any implementation which
+does this will be extremely susceptible to DoS attacks because UDP forgery
+is so easy.
```
ekr, line 895: I think we can actually remove this paragraph because all TLS 1.3 alerts are fatal.

draft-ietf-tls-rfc9147bis.md
```diff
@@ -845,25 +845,59 @@ record has been deprotected  successfully.
...
+
+Note that this means that DTLS does not safeguard against denial of service
+caused by damage to handshake messages carried in unprotected records.
+
+In general, invalid records SHOULD be silently discarded, thus preserving
+the association; however, an error MAY be logged for diagnostic purposes.
+
+Generating fatal alerts is NOT RECOMMENDED for unreliable transports, both to
+increase the reliability of DTLS service and to avoid the risk of spoofing
+attacks sending traffic to unrelated third parties.
```
ekr: I think we can actually remove this paragraph because all TLS 1.3 alerts are fatal.

ekr (changes requested):

draft-ietf-tls-rfc9147bis.md
martinthomson:
```suggestion
```

draft-ietf-tls-rfc9147bis.md
martinthomson: Is it worth saying something like this (perhaps above)?

> Any implementation that treats deprotection [this is not a word dammit] failures as fatal will be highly susceptible to DoS, because UDP forgery is so easy.

ekr:
Superseded by #330

Issue #325, opened 2026-06-28
Clarify when receivers close old epochs


hannestschofenig:
Errata 8108 fixed one concrete ACK validation problem: an ACK must not reference record numbers from an epoch higher than the epoch in which the ACK was received.

David Benjamin's errata note also raised a broader question:

> Otherwise we need to be much more explicit about the points at which the
> receiver MUST close old epochs. Honestly, we probably should be explicit
> about this too, but we should also be clear on this point.

The current draft says that implementations SHOULD discard records from earlier epochs, but MAY retain keying material from previous epochs for reordering. This does not clearly define when a receiver MUST stop accepting records from an old epoch and discard the corresponding receive-side keying material/replay state.

Does DTLS 1.3-bis needs explicit epoch closure points? (including for the handshake epochs, 0-RTT epoch, application data epochs, and KeyUpdate-generated epochs)

PR #326, opened 2026-06-28
Clarify receive epoch closure

hannestschofenig:
Aiming to address this issue: https://github.com/tlswg/dtls13-spec/issues/325

This PR clarifies when DTLS 1.3 receivers may close old receive epochs.

The text attempts to define what it means to close a receive epoch and explains that, in the base protocol, receive epochs are created by the initial handshake and, after the handshake, by KeyUpdate. Other post-handshake messages are protected under the current application traffic epoch and do not create a new epoch.

### DIFF
--- a/draft-ietf-tls-rfc9147bis.md
+++ b/draft-ietf-tls-rfc9147bis.md
@@ -582,6 +582,48 @@ the IETF for MSL, as specified in {{RFC0793}} or successors,
 not that they attempt to interrogate the MSL that
 the system TCP stack is using.)
 
+For the purposes of this section, closing a receive epoch means discarding
+the record-protection keys, record number reconstruction state, replay state,
+and any ACK processing state associated with that epoch. Once a receive epoch
+is closed, records from that epoch MUST be treated as records for which no
+corresponding cipher state can be determined.
+
+In the base protocol, receive epochs are created by the initial handshake and,
+after the handshake has completed, by KeyUpdate. Other post-handshake messages
+are protected under the current application traffic epoch and do not create a
+new epoch.
+
+During the initial handshake, a receiver MAY retain receive state for older
+epochs to tolerate packet reordering or to complete retransmission and
+acknowledgement processing. Closing an epoch affects only records from that
+epoch that arrive after the closure; those records will be discarded as
+undecryptable. A receiver MAY close epoch 0 after it has processed the
+plaintext handshake messages needed to establish the next receive epoch and no
+longer needs to process retransmissions of those messages. A server MAY close
+epoch 1 when it rejects early data or when it no longer intends to accept
+reordered 0-RTT records. A receiver MAY close epoch 2 after the handshake has
+completed and ACK or retransmission processing for records protected under the
+handshake traffic keys has completed or been abandoned.
+
+After the handshake, the base protocol creates new receive epochs only through
+KeyUpdate. A receiver MAY retain the previous application traffic epoch to
+tolerate reordering and to process records, including post-handshake records,
+that were already sent under that epoch. Closing an older application traffic
+epoch too early can therefore cause loss of reordered application data, prevent
+processing of post-handshake records protected under that epoch, or violate the
+KeyUpdate retention requirements below. A receiver MAY close an older
+application traffic epoch when it no longer intends to tolerate reordering from
+that epoch, subject to the KeyUpdate rules below and any extension-specific
+rules.
+
+In particular, a receiver MUST close an older receive epoch once it is no longer
+needed for reordering, retransmission, or acknowledgement processing, and
+MUST NOT retain receive state for an older epoch longer than the default MSL.
+Protected DTLS 1.3 records carry only the low-order bits of the epoch.
+A receiver MUST close older receive epochs before retaining a later epoch would
+make epoch reconstruction ambiguous. In particular, a receiver MUST NOT retain
+two receive epochs with the same encoded epoch bits for the same peer.
+
 Conversely, it is possible for records that are protected with the
 new epoch to be received prior to the completion of a
 handshake.  For instance, the server may send its Finished message
@@ -1317,7 +1359,14 @@ from the wire and the handshake transcript. Because DTLS
 records have epochs, EndOfEarlyData is not necessary to determine
 when the early data is complete, and because DTLS is lossy,
 attackers can trivially mount the deletion attacks that EndOfEarlyData
-prevents in TLS. Servers SHOULD NOT accept records from epoch 1 indefinitely once they are able to process records from epoch 3. Though reordering of IP packets can result in records from epoch 1 arriving after records from epoch 3, this is not likely to persist for very long relative to the round trip time. Servers could discard epoch 1  keys after the first epoch 3 data arrives, or retain keys for processing epoch 1 data for a short period.
+prevents in TLS. Servers SHOULD NOT accept records from epoch 1 indefinitely
+once they are able to process records from epoch 3. Though reordering of IP
+packets can result in records from epoch 1 arriving after records from epoch 3,
+this is not likely to persist for very long relative to the round trip time.
+Servers could discard epoch 1 keys after the first epoch 3 data arrives, or
+retain keys for processing epoch 1 data for a short period. In all cases,
+servers MUST close epoch 1 once they no longer need it for reordering, and MUST
+NOT retain epoch 1 longer than the default MSL.
 (See {{dtls-epoch}} for the definitions of each epoch.)
 
 
@@ -2101,7 +2150,10 @@ of the NewSessionTicket indefinitely until its maximum retransmission count is r
 # Key Updates
 
 As with TLS 1.3, DTLS 1.3 implementations send a KeyUpdate message to
-indicate that they are updating their sending keys.  As with other
+indicate that they are updating their sending keys. In the base protocol,
+KeyUpdate is the only post-handshake message that creates a new application
+traffic epoch; other post-handshake messages are protected under the current
+application traffic epoch. As with other
 handshake messages with no built-in response, KeyUpdates MUST be
 acknowledged.  In order to facilitate epoch reconstruction
 ({{reconstructing}}), implementations MUST NOT send records with the new keys or
@@ -2119,7 +2171,13 @@ such out-of-epoch records.
 Due to the possibility of an ACK message for a KeyUpdate being lost and thereby
 preventing the sender of the KeyUpdate from updating its keying material,
 receivers MUST retain the pre-update keying material until receipt and successful
-decryption of a message using the new keys.
+decryption of a message using the new keys. After successfully decrypting a
+message using the new keys, receivers MUST close the previous receive epoch
+once it is no longer needed for reordering or retransmission processing, as
+described in {{seq-and-epoch}}. Extensions that define a different key update
+state machine, such as Extended Key Update {{?I-D.ietf-tls-extended-key-update}}, are
+responsible for defining the receive epoch retention and closure points for
+that state machine.
 
 {{dtls-key-update}} shows an example exchange illustrating that successful
 ACK processing updates the keys of the KeyUpdate message sender, which is
--- a/draft-ietf-tls-rfc9147bis.md
+++ b/draft-ietf-tls-rfc9147bis.md
@@ -571,16 +571,17 @@ are provided in {{dtls-epoch}}.
 
 ### Processing Guidelines
 
-Because DTLS records could be reordered, a record from epoch
-M may be received after epoch N (where N > M) has begun.
-Implementations SHOULD discard records from earlier epochs but
-MAY choose to
-retain keying material from previous epochs for up to the default MSL
-specified for TCP {{RFC0793}} to allow for packet reordering.  (Note that
-the intention here is that implementers use the current guidance from
-the IETF for MSL, as specified in {{RFC0793}} or successors,
-not that they attempt to interrogate the MSL that
-the system TCP stack is using.)
+Because DTLS records can arrive out of order, a record from epoch
+M might be received after epoch N, where N > M, has already begun.
+Implementations SHOULD discard records from earlier epochs. However,
+they MAY retain keying material for previous epochs for up to the
+default Maximum Segment Lifetime (MSL) specified for TCP in {{RFC0793}}
+or its successors, in order to tolerate packet reordering.
+
+The intent is that implementers follow the current IETF guidance for
+TCP MSL, as specified in {{RFC0793}} or successor documents.
+Implementations are not expected to determine or depend on the MSL
+value used by the local system TCP stack.
 
 For the purposes of this section, closing a receive epoch means discarding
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc9147bis.md
```diff
@@ -582,6 +582,48 @@ the IETF for MSL, as specified in {{RFC0793}} or successors,
...
+epoch too early can therefore cause loss of reordered application data, prevent
+processing of post-handshake records protected under that epoch, or violate the
+KeyUpdate retention requirements below. A receiver MAY close an older
+application traffic epoch when it no longer intends to tolerate reordering from
+that epoch, subject to the KeyUpdate rules below and any extension-specific
+rules.
+
+In particular, a receiver MUST close an older receive epoch once it is no longer
+needed for reordering, retransmission, or acknowledgement processing, and
+MUST NOT retain receive state for an older epoch longer than the default MSL.
```
Frosne, line 622: Could you elaborate on MSL? 

Default MSL is a) TCP one (2 minutes) b) an implementation defined one? c) something else we need to define?

hannestschofenig: The MSL from RFC 793 is 2 minutes. Further up in the section we say the following:

>    (Note
>    that the intention here is that implementers use the current guidance
>    from the IETF for MSL, as specified in [RFC0793] or successors, not
>    that they attempt to interrogate the MSL that the system TCP stack is
>    using.)

I have updated the first two paragraphs to increase visibility. Let me know what you think about the update.

hannestschofenig:
I made further changes to the text to be more precise about when the epoch can be closed in case of a key update.

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+epochs to tolerate packet reordering or to complete retransmission and
+acknowledgement processing. Closing an epoch affects only records from that
+epoch that arrive after the closure; those records will be discarded as
+undecryptable. A receiver MAY close epoch 0 after it has processed the
+plaintext handshake messages needed to establish the next receive epoch and no
+longer needs to process retransmissions of those messages. A server MAY close
+epoch 1 when it rejects early data or when it no longer intends to accept
+reordered 0-RTT records. A receiver MAY close epoch 2 after the handshake has
+completed and ACK or retransmission processing for records protected under the
+handshake traffic keys has completed or been abandoned.
```
ekr, line 607: I think we may need to explain this more, or point upwards:

1. A client can close epoch 2 after it has received an ACK for Finished.
2. A server can close epoch 2 2MSL after receiving Finished.

This is implicit in 5.8.1. Right?

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+epoch too early can therefore cause loss of reordered application data, prevent
+processing of post-handshake records protected under that epoch, or violate the
+KeyUpdate retention requirements below. A receiver MAY close an older
+application traffic epoch when it no longer intends to tolerate reordering from
+that epoch, subject to the KeyUpdate rules below and any extension-specific
+rules.
+
+In particular, a receiver MUST close an older receive epoch once it is no longer
+needed for reordering, retransmission, or acknowledgement processing, and
+MUST NOT retain receive state for an older epoch longer than the default MSL.
```
ekr, line 622: What is this graf going that is not stated above?

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+Protected DTLS 1.3 records carry only the low-order bits of the epoch.
+A receiver MUST close older receive epochs before retaining a later epoch would
+make epoch reconstruction ambiguous. In particular, a receiver MUST NOT retain
+two receive epochs with the same encoded epoch bits for the same peer.
```
ekr, line 626: Why is this necessary? It seems like 4.2.2 lets you retain a longer period if you are willing to do trial decryption.

draft-ietf-tls-rfc9147bis.md
```diff
@@ -1317,7 +1360,14 @@ from the wire and the handshake transcript. Because DTLS
...
 attackers can trivially mount the deletion attacks that EndOfEarlyData
-prevents in TLS. Servers SHOULD NOT accept records from epoch 1 indefinitely once they are able to process records from epoch 3. Though reordering of IP packets can result in records from epoch 1 arriving after records from epoch 3, this is not likely to persist for very long relative to the round trip time. Servers could discard epoch 1  keys after the first epoch 3 data arrives, or retain keys for processing epoch 1 data for a short period.
+prevents in TLS. Servers SHOULD NOT accept records from epoch 1 indefinitely
+once they are able to process records from epoch 3. Though reordering of IP
+packets can result in records from epoch 1 arriving after records from epoch 3,
+this is not likely to persist for very long relative to the round trip time.
+Servers could discard epoch 1 keys after the first epoch 3 data arrives, or
+retain keys for processing epoch 1 data for a short period. In all cases,
+servers MUST close epoch 1 once they no longer need it for reordering, and MUST
+NOT retain epoch 1 longer than the default MSL.
```
ekr, line 1370: This also seems redundant. Is there a reason we need it?

ekr (changes requested):
I'm not sure this is right. I see David's erratum but I guess I'm not clear on why we need to define MUST's here. @davidben can you elaborate a bit?

draft-ietf-tls-rfc9147bis.md
hannestschofenig, line 607: Yes, this is implicit in Section 5.8.1. It is, however, useful to describe this explicitly. How about the following text:

```suggestion
undecryptable. For the final handshake flight, the retransmission state machine in
{{state-machine}} determines when this processing is complete. A client that
has sent Finished can close epoch 2 after it has received an ACK for that
Finished, because the ACK for the final flight cancels retransmission
processing. A server that has received the client's Finished still needs to
retain epoch 2 receive state for at least twice the default MSL so that it can
process retransmissions of the client's final flight and retransmit its ACK;
the server can close epoch 2 after that period has expired.
```

draft-ietf-tls-rfc9147bis.md
hannestschofenig, line 622: You are right. The text does not really much value. It repeats text.

```suggestion
```

draft-ietf-tls-rfc9147bis.md
hannestschofenig, line 626: I have updated the text and included a reference to Section 4.2.2:

```suggestion
Protected DTLS 1.3 records carry only the low-order bits of the epoch.
Retaining multiple receive epochs with the same encoded epoch bits can make
epoch reconstruction ambiguous. Implementations that retain such epochs need to
account for this when reconstructing the full epoch; see {{reconstructing}}.
```

draft-ietf-tls-rfc9147bis.md
hannestschofenig, line 1370: I agree. I went a bit overboard with the explanation.

```suggestion
prevents in TLS. Servers that process 0-RTT data can therefore close epoch 1 when
they no longer intend to accept reordered 0-RTT records, as described in
{{seq-and-epoch}}.
```

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+M might be received after epoch N, where N > M, has already begun.
+Implementations SHOULD discard records from earlier epochs. However,
+they MAY retain keying material for previous epochs for up to the
+default Maximum Segment Lifetime (MSL) specified for TCP in {{RFC0793}}
+or its successors, in order to tolerate packet reordering.
+
+The intent is that implementers follow the current IETF guidance for
+TCP MSL, as specified in {{RFC0793}} or successor documents.
+Implementations are not expected to determine or depend on the MSL
+value used by the local system TCP stack.
```
hannestschofenig, line 584:
```suggestion
Implementations SHOULD discard records from earlier epochs. However, they MAY
retain receive state from previous epochs temporarily to tolerate packet
reordering or to complete retransmission and acknowledgement processing, as
described below.

Where this section refers to the default MSL, the intention is that implementers
use the current guidance from the IETF for MSL, as specified in {{RFC0793}} or
successor documents. Implementations are not expected to determine or depend on
the MSL value used by the local system TCP stack.
```

hannestschofenig: Text update to refer to "receive state" rather than "keying material" since the receive contains more than just keying material. I think this is more precise.

Then, I avoid talking about MSL here twice since we have the detailed guidance in subsequent paragraphs anyway.

Frosne: I'd not talk about the intention here.. I'd have smth like: 

In this section, the default MSL is the value recommended by current IETF guidance, as specified in {{RFC0793}} or a successor document. An implementation does not need to determine, and must not depend on, the MSL value used by the local TCP stack.

draft-ietf-tls-rfc9147bis.md
Frosne: After this period (of MSL) expires, the server MAY discard the epoch 2 receive state.?

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+is closed, records from that epoch MUST be treated as records for which no
+corresponding cipher state can be determined.
+
+In the base protocol, receive epochs are created by the initial handshake and,
+after the handshake has completed, by KeyUpdate. Other post-handshake messages
+are protected under the current application traffic epoch and do not create a
+new epoch.
+
+During the initial handshake, a receiver MAY retain receive state for older
+epochs to tolerate packet reordering or to complete retransmission and
```
davidben, line 598: This sentence is a little odd to me, given there's another sentence at the start of the section that says basically the same thing.

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+acknowledgement processing. Closing an epoch affects only records from that
+epoch that arrive after the closure; those records will be discarded as
+undecryptable. A receiver MAY close epoch 0 after it has processed the
+plaintext handshake messages needed to establish the next receive epoch and no
+longer needs to process retransmissions of those messages. A server MAY close
```
davidben, line 603: This sentence threw me off a bit, but I may have just read it funny. You could read this sentence in two ways:

1. A receiver MAY close epoch 0 {after it has processed the plaintext handshake messages needed toestablish the next receive epoch} and no longer needs to process retransmissions of those messages
2. A receiver MAY close epoch 0 after {it has processed the plaintext handshake messages needed to establish the next receive epoch and no longer needs to process retransmissions of those messages}

For some reason, my mind went to the first interpretation, but the first interpretation is not true. A server, after it sends ServerHello, needs to retain the epoch 0 *at least* until it has received one epoch 1 record. At that point, it can assume the client has received ServerHello.

The second interpretation is true but not super helpful unless you know when that point happens.

I do worry that all these handshake-state specific rules make the protocol a little fussy to implement. In the thread below I sketched out what we do. At least, I think it's what we do. This was years ago and I've honestly forgotten a lot of it. :-)

draft-ietf-tls-rfc9147bis.md
davidben: I agree this is worth calling out. Case 2 was something we had to fill in after our initial implementation.

In our implementation, we tried to go with a straightforward rule, that required less syncing between the handshake state and the record layer: when you open a new receive epoch N, you don't immediately install it. You instead put it in the "next epoch" slot. As soon as you receive a record from that epoch, you immediately promote it to the current epoch. At this point, you know the other side is capable of sending at this epoch and shouldn't need to care about the previous epoch anymore.

The problem being exactly case 2. As a server, you might well receive epoch 3 from the client and say, oh, I guess we can close epoch 2 now. But the client hasn't received your ACK for server finished and is sitting there retransmitting, wait for you to ACK it. So epoch 2 needs to remain open until then.

So what we ended up doing was to have a conveyor belt of three slots. prev/current/next:
* You're allowed to receive from any of those three epochs
* New epochs go in `next` when they're opened
* If you receive an epoch at `next`, the conveyor belt instantly shifts forward. `prev` falls off, `current` becomes `prev`, and `next` becomes `current`
* `prev` lives for two MSL and then gets closed on a timer

This seems to have worked OK for all the cases that have come up so far? We do end up keeping epoch 2 alive on the client for slightly longer than strictly necessary, but that seems OK.

(I suspect we could have gotten away with only two of them and removed the `next` state. We initially only had `current` and `next` and only added `prev` when we realized this extra case. But with two of them, you need to track extra state on `prev` to know whether it's dying on a timer, or kept strictly alive until you've gotten anything on the new epoch.)

NB: We didn't implement 0-RTT. When/if we do, we'll have to teach it that epoch 1 is out of sequence with epoch 2 and managed separately. I have to page all this back in, but I think the behavior I'd do is something like... once you receive something from epoch 3, you put epoch 1 on a 2 MSL timer (as if it were in `prev`) and let it expire that way.

draft-ietf-tls-rfc9147bis.md
```diff
@@ -571,16 +571,59 @@ are provided in {{dtls-epoch}}.
...
+epoch 1 when it rejects early data or when it no longer intends to accept
+reordered 0-RTT records. A receiver MAY close epoch 2 after the handshake has
+completed and ACK or retransmission processing for records protected under the
+handshake traffic keys has completed or been abandoned.
+
+After the handshake, the base protocol creates new receive epochs only through
+KeyUpdate. A receiver MAY retain the previous application traffic epoch to
```
davidben, line 610: This is a bit subtle. In some cases this MAY is a MUST. When you send KeyUpdate, you immediately *create* the new receive epoch, but it's not allowed to supplant the previous one until you've received *something* on the new receive epoch. From Section 8:

> Due to the possibility of an ACK message for a KeyUpdate being lost and thereby preventing the sender of the KeyUpdate from updating its keying material, receivers MUST retain the pre-update keying material until receipt and successful decryption of a message using the new keys

(This was where we got our `next` vs `current` model from, and noticing that it worked alright for the handshake too.)

draft-ietf-tls-rfc9147bis.md
davidben: I agree with removing this. The lifecycle of epochs in DTLS 1.3 is complex enough that I think we reasonably have to allow implementations to apply more generic rules. Indeed I think one of the problems here is that the spec makes it *really* hard to reason about which generic rules are and aren't safe.

davidben (commented):
Finally got to looking at this. Sorry it took so long.

I think my high-level comment is it would be nice if we could make epoch management a bit less complex. I.e. not requiring such tight coordination between the handshake state.

